PR #13, opened 2022-02-17
Dtype itemsize

jperez999:
hope this fits.

### DIFF
--- a/merlin/schema/io/tensorflow_metadata.py
+++ b/merlin/schema/io/tensorflow_metadata.py
@@ -240,6 +240,7 @@ def extract_merlin_properties(feature):
 
     return properties
 
+
 def merlin_properties(feature):
     properties = extract_merlin_properties(feature)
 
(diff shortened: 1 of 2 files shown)


merlin/schema/io/tensorflow_metadata.py
```diff
@@ -14,6 +14,7 @@
 #
 import pathlib
 from typing import Union
+import warnings
```
benfred, line 17: I think this import is unused  

```suggestion
```

merlin/schema/io/tensorflow_metadata.py
```diff
@@ -245,17 +253,37 @@ def merlin_properties(feature):
...
+float_dtypes_map = {
+    16: numpy.float16,
+    32: numpy.float32,
+    64: numpy.float64,
+}
 
 
 def merlin_dtype(feature):
     dtype = None
+    item_size = int(extract_merlin_properties(feature).get("dtype_item_size", 0)) or None
```
benfred, line 276: Rather than extract twice - can we pass the properties extracted by merlin_properties to merlin_dtype? 

Also can we pop the dtype_item_size value from the properties (like we do from is_ragged/is_list) when converting back a merlin schema object?

benfred (commented):
Looks good! thanks for this.  

I think CI didn't kick off because the repo was private and this PR is submitted on a fork.  I think CI would have complained about an unused import though - can you install the pre-commit hook and run on this ? (```pre-commit run --all-files```) . I've made the repo public now so CI should kick in after any changes.

PR #35, opened 2022-03-07
Add docstrings to `merlin.schema`

karlhigley:
Bumps docstring coverage requirement to 50%

### DIFF
--- a/merlin/schema/io/proto_utils.py
+++ b/merlin/schema/io/proto_utils.py
@@ -25,12 +25,22 @@ ProtoMessageType = TypeVar("ProtoMessageType", bound=BetterProtoMessage)
 def has_field(message: ProtoMessageType, field_name: str) -> bool:
     """Check if a Protobuf message has a particular field
 
-    Args:
-        message (ProtoMessageType): Protobuf message object
-        field_name (str): Name of the field to look for
+    Parameters
+    ----------
+    message : ProtoMessageType
+        Protobuf message object
+    field_name : str
+        Name of the field to look for
+    message: ProtoMessageType :
 
-    Returns:
-        bool: Whether or not the named field exists on the message object
+    field_name: str :
+
+
+    Returns
+    -------
+    bool
+        Whether or not the named field exists on the message object
+
     """
     return betterproto.serialized_on_wire(getattr(message, field_name))
 
@@ -38,11 +48,16 @@ def has_field(message: ProtoMessageType, field_name: str) -> bool:
 def copy_better_proto_message(better_proto_message: ProtoMessageType, **kwargs) -> ProtoMessageType:
     """Create a copy of a Protobuf message
 
-    Args:
-        better_proto_message (ProtoMessageType): The message to copy
+    Parameters
+    ----------
+    better_proto_message : ProtoMessageType
+        The message to copy
 
-    Returns:
-        ProtoMessageType: Copy of better_proto_message
+    Returns
+    -------
+    ProtoMessageType
+        Copy of better_proto_message
+
     """
     output = better_proto_message.__class__().parse(bytes(better_proto_message))
     for key, val in kwargs.items():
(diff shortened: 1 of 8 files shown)


merlin/schema/io/tensorflow_metadata.py
```diff
@@ -35,69 +35,138 @@
...
+        """Convert this TensorflowMetadata schema object to a JSON string
+
+        Returns:
+            str: Schema converted to a JSON string
+        """
         return self.proto_schema.to_json()
 
 
-def pb_int_domain(column_schema):
+def _pb_int_domain(column_schema):
```
karlhigley, line 169: Semi-private methods that start with a leading underscore are exempted from the docstring requirement, and these don't appear to be called from anywhere else so I made them semi-private.

(Over the long haul, we may also want to add docstrings to semi-private methods, but trying to keep the scope manageable for 22.04.)

merlin/schema/io/proto_utils.py
```diff
@@ -22,11 +22,43 @@
...
+        Name of the field to look for
+    message: ProtoMessageType :
+
+    field_name: str :
+
+
+    Returns
+    -------
+    bool
+        Whether or not the named field exists on the message object
```
mikemckiernan, line 42: This might sound pithy, but can you be even more explicit and include `true` or `false`?

> `True` if the named field exists on the message object

merlin/schema/io/tensorflow_metadata.py
```diff
@@ -35,69 +35,197 @@
...
+        Parameters
+        ----------
+        path : str
+            Path to the directory containing the Protobuf text file
+        file_name : str
+            Name of the output file. Defaults to "schema.pbtxt".
+        path: str :
+
+
+        Returns
```
mikemckiernan, line 149: Sounds like standards are in flight.  I think Ben suggested omission.  Please let me know either way.  We need it in the CONTRIBUTING doc.

merlin/schema/schema.py
```diff
@@ -40,6 +40,11 @@ class ColumnSchema:
     is_ragged: bool = False
 
     def __post_init__(self):
+        """Standardize tags and dtypes on initialization
+
+        Raises:
+            TypeError: If provided dtype can't be cast to a numpy dtype
```
mikemckiernan, line 46: Up to you: I'd spell out "cannot" and I'd include "the" to avoid telegraphic language:

> TypeError: If the provided dtype cannot be cast to a numpy dtype

mikemckiernan (commented):
I don't want to hold you up, so I identified the few items that I felt like I had some footing to defend.

merlin/schema/schema.py
```diff
@@ -223,9 +370,42 @@ def without(self, col_names):
...
+    def first(self) -> ColumnSchema:
+        """Returns the first ColumnSchema in the Schema. Useful for cases where you select down
+        to a single column via select_by_name or select_by_tag, and just want the value
+
+        Parameters
+        ----------
+
+        Returns
+        -------
+
```
benfred, line 402: I think we can remove Parameters here - but we probably should include the Returns type (rather than leaving blank)

merlin/schema/io/tensorflow_metadata.py
```diff
@@ -35,69 +35,197 @@
 
 
 class TensorflowMetadata:
+    """Reads and writes Merlin schemas as Protobuf and JSON files"""
```
benfred, line 38: minor nitpick: I think we could expand on this - since we're not just reading reading and writing schemas as Protobuf/JSON, but specifically we're formatting as the format from the  https://github.com/tensorflow/metadata project. Maybe we can add an extra sentence with a link here?

benfred (commented):
A couple of minor things - but looks good! Thanks for adding the extra doc strings

merlin/schema/io/tensorflow_metadata.py
benfred: I think thats a good callout, and we should get in the habit of separating out public/private API's

benfred (approved):
lgtm! thanks

PR #36, opened 2022-03-09
chore(docs): configure doc preview for PRs

mikemckiernan:
Complete the configuration for a Sphinx build.

Add the workflow for doc previews in PRs.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -31,7 +31,6 @@ here](https://nvidia-merlin.github.io/core).
    directory:
 
    ```sh
-   http-server build/html
+   python -m http.server -d build/html
    ```
 
-   (http-server can be installed with `npm install --global http-server`)
--- a/.github/workflows/cpu-ci.yml
+++ b/.github/workflows/cpu-ci.yml
@@ -8,7 +8,7 @@ on:
       - v*
   pull_request:
     branches: [ main ]
-
+    types: [opened, synchronize, reopened, closed]
 
 jobs:
   build:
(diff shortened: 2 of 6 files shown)


mikemckiernan:
As with the previous two times, I don't anticipate that the `docs-preview-pr` workflow will run.

In terms of a check, I pushed this commit to the `main` branch of my fork and then opened a PR against it.  It seems OK:

<https://github.com/mikemckiernan/core/pull/2>

docs/README.md
```diff
@@ -31,7 +31,7 @@ here](https://nvidia-merlin.github.io/core).
    directory:
 
    ```sh
-   http-server build/html
+   python -m http.server -d build/html
```
benfred, line 34: the python http.server module isn't the best imo (and I prefer using the node.js version that was previously listed here - it handles concurrent requests much better, and doesn't flake out as much at least for me).

However, If we are going to recommend going with the python http.server, maybe we can ditch the recommendation on line 37 to install the node.js one ?

benfred (approved):
thanks! looks good

I'll create the gh-pages branch in a second

docs/README.md
mikemckiernan: Doh!  I missed that.  I just noticed that I wasn't successful running the command so I substituted back in the vanilla Python method.

I'm not sure that my opinion matters for much, but to me, I have a preference to keeping all the dependencies for a project in a distinct venv.  Mixing in node...I guess my pref is to show the simplest way.  More knowledgeable folks like yourself don't have to follow instructions.

I'll remove line 37. Thank you.

PR #53, opened 2022-03-31
README updates

benfred:
* Add badges
* Add Feedback/support section
* Add docker install instructions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,7 @@ cover/
 
 # Sphinx documentation
 docs/build/
+docs/source/README.md
 
 # IPython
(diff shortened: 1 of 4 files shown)


mikemckiernan:
@benfred , do you mind if I take this PR and branch? I'll work the README into an introduction for the doc.  I suspect that I'll find something nitpicky too.

Oh, for example, I'm inclined to use "Merlin Core" when the words are intended for humans and `merlin-core` when the intended audience is `pip` or (I guess...) `conda`.

karlhigley (approved):

PR #56, opened 2022-04-01
Exclude `merlin/io` from docstring coverage checks

karlhigley:
This module only has one public class (`Dataset`), which is already well-documented.


viswa-nvidia:
NOt a blocker

benfred (approved):

PR #65, opened 2022-04-05
Remove use of `kind` property in `ColumnSchema`

karlhigley:
This was intended to normalize Pandas nullable types to their closest corresponding Numpy dtypes, but turns out to break with other numpy dtypes that also define the `kind` property, so must be reverted.

### DIFF
--- a/merlin/schema/schema.py
+++ b/merlin/schema/schema.py
@@ -53,8 +53,6 @@ class ColumnSchema:
                 dtype = np.dtype(self.dtype.numpy_dtype)
             elif hasattr(self.dtype, "_categories"):
                 dtype = self.dtype._categories.dtype
-            elif hasattr(self.dtype, "kind"):
-                dtype = np.dtype(self.dtype.kind)
             else:
                 dtype = np.dtype(self.dtype)
         except TypeError as err:


merlin/schema/schema.py
```diff
@@ -53,8 +53,6 @@ def __post_init__(self):
                 dtype = np.dtype(self.dtype.numpy_dtype)
             elif hasattr(self.dtype, "_categories"):
                 dtype = self.dtype._categories.dtype
-            elif hasattr(self.dtype, "kind"):
-                dtype = np.dtype(self.dtype.kind)
```
rjzamora, line 57: I think we still need to handle string dtypes (the parquet change only handles int/float nullable dtypes):

```python
            elif isinstance(self.dtype, pd.StringDtype):
                dtype = np.dtype("O")
```

rjzamora: To clarify - We *could* add a check for pandas string types in the parquet reader, but I believe they are created in other places anyway...

rjzamora: We avoid creating nullable integer and float dtypes in parquet now, but we weren't creating `StringDtype` there anyway (I think - I need to check that).

rjzamora: Okay - I think you are right. I missed that some categorical columns are producing StringDtype - I'll try another tweak to the parquet reader and see if we can just drop the block as you are suggesting here.

rjzamora: Okay - It does seem that your original change here works fine if #66 is also merged :)

rjzamora:
LGTM, but it looks like you will need someone with higher merlin privileges to approve as well :)

PR #88, opened 2022-05-12
chore: Add release-drafter

mikemckiernan:
Introduce use of the release-drafter
action to the repository. This addition
should help us with creating release notes.


.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,23 @@
+name: Release Drafter
```
benfred, line 1: did you mean to include this file? All the other PR's are just the release-drafter config file, and not the workflow

mikemckiernan: Removed.  I opened a few PRs before remembering the organizational-level approach.

benfred (approved):

PR #91, opened 2022-05-18
Add `MerlinArray` for conversion between TF, Numpy, CuPy, and CuDF

karlhigley:


jperez999 (approved):

jperez999:
we need to add test skips if we don't have the target package in environment.

jperez999:
rerun tests

PR #92, opened 2022-05-19
Adds a common triage workflow

benfred:
Use a reusable workflow for triaging issues, rather than have a bunch of
boilerplate defined here.


mikemckiernan (approved):
/lgtm

PR #94, opened 2022-05-20
Add predicate function argument to `select_by_tag` 

oliverholworthy:
### Goals :soccer:

When using the `select_by_tag` method on a schema. If passing multiple tags to the method. Enable column selection by tag with both AND (`all`) and OR (`any`) logic. The current default is OR (`any`). When providing multiple tags, the columns selected can may any of the tags provided. 

A motivating example for this is if we separate something like the `ITEM_ID` tag into two. Instead of performing `schema.select_by_tag(Tags.ITEM_ID)` you'd need to do something like `schema.select_by_tag([Tags.ITEM, Tags.ID])`. Currently, this would return all the columns tagged _ITEM_ and all columns tagged with _ID_, and not the item ID column.

Note that it is possible to achieve the selection of all tags with the current API by chaining calls to `select_by_tag`. 

```python
from merlin.schema import ColumnSchema, Schema

col1_schema = ColumnSchema("col1", tags=["a", "b"])
col2_schema = ColumnSchema("col2", tags=["a"])

schema = Schema([col1_schema, col2_schema])

schema.select_by_tag("a").select_by_tag("b")  # => returns col1
```

This change makes this more convenient by supporting a single method call to achieve this. The equivalent of the line above with this PR:

```
schema.select_by_tag(["a", "b"], all)  # => returns col1
```

### Implementation Details :construction:

This PR adds a new argument to `select_by_tag` that enables control over the predicate function used to determine column selection from tags.

Using the same example as above. The PR changes the result of `select_by_tag`:

#### Before

```python
schema.select_by_tag(["a", "b"])  # => returns col1 and col2
```

#### After

```python
# both equivalent to before
schema.select_by_tag(["a", "b"])       # => returns col1 and col2
schema.select_by_tag(["a", "b"], any)  # => returns col1 and col2

# changing predicate from any to all
schema.select_by_tag(["a", "b"], all)  # => returns col1
```

We may wish to swap the default value around from any to all in a further PR, however that would be a breaking change and will be clearer to separate in another PR


tests/unit/schema/test_schema.py
```diff
@@ -51,13 +52,35 @@ def test_select_by_tag():
...
 
+def test_select_by_any_tag():
+    col1_schema = ColumnSchema("col1", tags=["a", "b", "c"])
+    col2_schema = ColumnSchema("col2", tags=["b", "c", "d"])
+    col3_schema = ColumnSchema("col3", tags=["b", "e", "f"])
+
+    schema = Schema([col1_schema, col2_schema, col3_schema])
+
+    col1_selection = schema.select_by_tag("a")
+    col2_selection = schema.select_by_tag("d")
```
karlhigley, line 70: This should probably either be updated to `select_by_any_tag` or dropped (since it's repeating a test that already exists elsewhere.)

merlin/schema/schema.py
```diff
@@ -286,7 +286,32 @@ def apply_inverse(self, selector) -> "Schema":
         return self.excluding(selector)
 
     def select_by_tag(self, tags: Union[Union[str, Tags], List[Union[str, Tags]]]) -> "Schema":
-        """Select matching columns from this Schema object using a list of tags
+        """Select all matching columns from this Schema object using a list of tags
```
karlhigley, line 289: These docstrings might need an update to capture the difference between matching any tag and matching all tags. Maybe something like "Select columns from this Schema object that match all supplied tags"?

merlin/schema/schema.py
```diff
@@ -286,7 +286,32 @@ def apply_inverse(self, selector) -> "Schema":
...
+
+        selected_schemas = {}
+
+        for _, column_schema in self.column_schemas.items():
+            if all(x in column_schema.tags for x in tags):
+                selected_schemas[column_schema.name] = column_schema
+
+        return Schema(selected_schemas)
+
+    def select_by_any_tag(self, tags: Union[Union[str, Tags], List[Union[str, Tags]]]) -> "Schema":
```
karlhigley, line 313: What would you think about making the methods `select_by_all_tags` and `select_by_any_tags`, and then adding a method alias that makes `select_by_tag` refer to `select_by_any_tags` for backwards compatibility? That might be smoother than changing the behavior of the existing method.

karlhigley:
Let's hold off on merging this until after the 22.08 release

nv-alaiacano:
Perhaps a non-breaking way to implement this would be to add an `all` parameter to `select_by_tag`, indicating that a column must match _all_ tags.

```python
schema.select_by_tag(["a", "b"])  # => returns col1 and col2
schema.select_by_tag(["a", "b"], all=True)  # => returns col1
```

karlhigley:
On reflection, I kinda like @nv-alaiacano's suggestion about adding a parameter to control whether it matches any or matches all. Normally, I'm not a fan of boolean arguments (which is probably why I suggested separate methods if I did in fact suggest that), but this seems like a case where it may be simpler/clearer to go that route. Another possibility would be to keep `select_by_tag` as is to avoid a breaking change and add a `select_by_all_tags` method. I could go either/any way here.

PR #102, opened 2022-06-15
Add Merlin Array classes

jperez999:
### MerlinArray
The main goal of this PR is to create a seamless experience in Merlin around data conversion. We would like to completely abstract converting data between frameworks (numpy, cupy, tensorflow, cudf, pandas). The user should not need to know anything about the data format and should not have to worry "is my data in the right format for this model". The MerlinArray class introduces a custom data representation class that tries to create new user selected views from original data with minimal copying (zero copy when possible). This concept introduces self registration which is required because of the dynamic environments in which our code is expected to operate (CPU or GPU).

### Virtual Dataframe / DataFrame Protocol
The Virtual DataFrame is a custom class created to house collections of merlin arrays. The Virtual dataframe is designed to take a collection of merlin array (of any types) and create a new view for a target framework i.e. `(VirtualDataFrame({"a": cupy.array([1,2,3,4])}).to_columns(np.array)` -> turn all merlin arrays in df into MerlinNumpyArrays) You can also create dataframe views using the `to` api. We also introduce the Dataframe protocol in this PR to allow for the interoperability of pandas/cudf dataframes and the Virtual DataFrame. The end design should allow for users to supply cudf, pandas, or virtual dataframes interchangeably to all operators. This sets the foundation for offline batch e2e processing.

### Features / Features 
Features lets you select the data (not just column names) according the supporting schema. Allowing you to select-by-tag and get back an object that has the data for those columns. This is different from current method because currently, select-by-tag gives you back a ColumnSelector with the columns you wanted and then you have to use those column names (from ColumnSelector) to access the data. Features, removes the intermediate steps and directly gives you the data for the tags requested.


merlin/features/array/cudf.py
```diff
@@ -0,0 +1,92 @@
...
+
+            Returns
+            -------
+            cudf.Series
+                The Series built from the array-like object
+            """
+            return cudf.io.from_dlpack(capsule)
+
+
+MerlinCudfArray = None if not cudf else _MerlinCudfArray
```
oliverholworthy, line 92: Could we have a class that raises on init instead of None, so that users get an error telling them that the corresponding package is not installed?

jperez999:
rerun tests

karlhigley:
Closing this since the approach with `@singledispatch` is preferred

PR #107, opened 2022-07-22
remove upstream dependencies that have no outputs

nv-alaiacano:
This resolves the bug reported in https://github.com/NVIDIA-Merlin/NVTabular/issues/1632

The issue was that we were successfully removing parents when calling `remove_inputs` but the node still showed up in the `dependencies` of the downstream nodes. We add logic to remove any dependency that does not have an `output_schema`.

I'll try adding a regression test as well.

### DIFF
--- a/merlin/dag/graph.py
+++ b/merlin/dag/graph.py
@@ -133,15 +133,21 @@ class Graph:
 
         while nodes_to_process:
             node, columns_to_remove = nodes_to_process.popleft()
-
             if node.input_schema and len(node.input_schema):
                 output_columns_to_remove = node.remove_inputs(columns_to_remove)
 
                 for child in node.children:
-                    nodes_to_process.append((child, to_remove + output_columns_to_remove))
+                    nodes_to_process.append(
+                        (child, list(set(to_remove + output_columns_to_remove)))
+                    )
 
                     if not len(node.input_schema):
                         node.remove_child(child)
+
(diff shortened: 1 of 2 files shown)


merlin/dag/graph.py
```diff
@@ -133,16 +133,22 @@ def remove_inputs(self, to_remove):
...
         while nodes_to_process:
             node, columns_to_remove = nodes_to_process.popleft()
-
             if node.input_schema and len(node.input_schema):
                 output_columns_to_remove = node.remove_inputs(columns_to_remove)
 
                 for child in node.children:
-                    nodes_to_process.append((child, to_remove + output_columns_to_remove))
+                    nodes_to_process.append(
+                        (child, list(set(to_remove + output_columns_to_remove)))
```
nv-alaiacano, line 141: This line's change is just to de-duplicate the list of output columns. without it, we would end up with something like `["target", "target"]` that would still technically work but made less sense when debugging this.

karlhigley:
LGTM once it has a test

benfred:
I just noticed this PR - after I had already submitted a fix of my own https://github.com/NVIDIA-Merlin/core/pull/108  =(

karlhigley (approved):

radekosmulski:
@benfred @nv-alaiacano @karlhigley I would like to thank you all for this super timely fix! 🙂

PR #108, opened 2022-07-23
Fix Graph.remove_inputs

benfred:
The `remove_inputs` call on the graph wasn't working - since it was hanging on
to a reference to the removed columns via the `dependencies` list. Fix and
add a unittest that would have caught this.

Fixes https://github.com/NVIDIA-Merlin/NVTabular/issues/1632

### DIFF
--- a/merlin/dag/node.py
+++ b/merlin/dag/node.py
@@ -150,6 +150,8 @@ class Node:
         for child_node in child_nodes:
             if self in child_node.parents:
                 child_node.parents.remove(self)
+            if self in child_node.dependencies:
+                child_node.dependencies.remove(self)
             if child_node in self.children:
                 self.children.remove(child_node)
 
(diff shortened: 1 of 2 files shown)


nv-alaiacano:
It seems like we were all working on this at the same time. Either PR is fine by me. https://github.com/NVIDIA-Merlin/core/pull/107

karlhigley (approved):

karlhigley:
I'm just noticing this PR after merging the other one, so...why not both? 😆

nv-alaiacano:
Closing this because #107 was merged and we don't want to merge both!

karlhigley:
I actually don't think it's problematic to merge both here, so long as the tests pass. There may be subsequent clean-up we want to do, but this fix is really nice.

PR #109, opened 2022-07-30
Add test for converting dataset to/from parquet

benfred:


benfred:
this was really just to test out https://github.com/NVIDIA-Merlin/core/issues/110 on CPU code (the existing `tests/unit/io/test_io.py` code is skipped on CPU right now) - but seems like this happens on its with GPU code too, so closing

PR #111, opened 2022-08-03
Update `value_count` serialization/deserialization to be consistent with original schema

oliverholworthy:
# Goal

- Ensure that the Schema (ColumnSchema) `value_count` property is restored to the same value when serialized/deserialized. 
- Update default value of `is_list`/`is_ragged` to infer from `value_count` property.

### `value_count` property serialization/deserialzation:

- fails to serialize to json
  - `value_count = {'max': 1, 'min': 1}`
  - `AttributeError: 'int' object has no attribute '_serialized_on_wire'`
- silently fails to re-construct the same schema object (serialize -> deserialize)
  - `value_count =` {`{'max': 0, 'min': 0}`, `{'max': 2, 'min': 1}`}

~These `value_count` examples may not be practically an issue~. It opens up some questions about where validation should take place for the Schema (at schema construction time or at serialization time). And where there are certain assumptions or constraints in serialation.

### `is_list`/`is_ragged` default value

- value_count max > min
  - implies that `is_ragged=True`, `is_list=True`
- value_count max == min
  - implies that `is_ragged=False`, `is_list=True`
- value_count max=0
  - implies that `is_ragged=False`, `is_list=False`


karlhigley (approved):

oliverholworthy:
This PR is a replacement for #169 since it removes the line being changed in that PR. #169 could be merged, but not strictly required to if this one is.

tests/unit/schema/test_column_schemas.py
```diff
@@ -180,8 +180,8 @@ def test_list_column_attributes():
     col2_schema = ColumnSchema("col2", is_list=True)
 
     assert col2_schema.is_list
-    assert col2_schema.is_ragged
-    assert col2_schema.quantity == ColumnQuantity.RAGGED_LIST
+    assert not col2_schema.is_ragged
+    assert col2_schema.quantity == ColumnQuantity.FIXED_LIST
```
oliverholworthy, line 184: This change reverts the default `is_ragged` value to `False` unless specified with the parameter or with a value count that has max > min.  

This was the case in merlin-core 0.5.0, but changed in 0.6.0 by #118

oliverholworthy: Changed this back to set the default value of is_ragged=True if is_list=True unless the value_count max > min

PR #118, opened 2022-08-18
Add a `quantity` attribute to `ColumnSchema`

karlhigley:
This is a step in the direction of safer and more consistent column schemas for list columns. This PR:

- Adds tests to the existing functionality and fixes a minor bug that occurs when the `is_list` flag is provided but the `is_ragged` flag is not
- Creates a `ColumnQuantity` enum with `SCALAR`/`FIXED_LIST`/`RAGGED_LIST` values and adds a property that returns the appropriate enum value
- Raises an error when attempting to create a schema with the invalid `is_list=False, is_ragged=True` state

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,14 +5,14 @@ repos:
     hooks:
       - id: absolufy-imports
   - repo: https://github.com/timothycrosley/isort
-    rev: 5.9.3
+    rev: 5.10.1
     hooks:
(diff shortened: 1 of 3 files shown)


nv-alaiacano (approved):

oliverholworthy (approved):
this should help simplify some checks in both systems and models and seems flexible enough that it doesn't tie us to a specific internal representation that we might want to change later

PR #119, opened 2022-08-18
Split compound tags (like `USER_ID`) into atomic tags (like `USER`,`ID`) 

karlhigley:
For now, this keeps the original compound tag in the list of tags in order to maintain backwards compatibility and allow us time to migrate our code, but eventually I think we can remove them.


oliverholworthy (approved):

merlin/schema/tags.py
```diff
@@ -26,30 +27,45 @@ class Tags(Enum):
...
     CONTEXT = "context"
 
     # Target related
     TARGET = "target"
-    BINARY_CLASSIFICATION = "binary_classification"
-    MULTI_CLASS_CLASSIFICATION = "multi_class"
     REGRESSION = "regression"
+    CLASSIFICATION = "classification"
+
+    BINARY = "binary"
```
marcromeyn, line 45: I am wondering if this (& `MULTI_CLASS`) could be infered automatically based on the int_domain?

oliverholworthy: could be represented outside of tags and instead with a property or method

would there ever be a use-case where we'd want to know if a column is binary or multi-class before it's been converted to an int? e.g. pre Categorify op, perhaps these are represented as strings.

karlhigley: It's possible that they could be inferred, but I think that's outside the scope of this PR (which is, after all, about making things more atomic) 😅

PR #121, opened 2022-08-19
Make the `Schema` class immutable

karlhigley:
An alternative to #120, which aims for increasing the degree of immutability (instead of decreasing it)


oliverholworthy (commented):
Since this is a breaking change to some of the uses in merlin models. Perhaps we should wait until we're sure the 22.08 release has been published along with the tagged version of the library before merging this PR. 

So that we have can make the required changes in the other libs in for the next release to use the immutable methods where required.

karlhigley:
I have updated to add the additional methods on `Schema` that make it possible to create new schema objects with modified column schemas

karlhigley:
rerun tests

merlin/schema/schema.py
```diff
@@ -426,6 +438,120 @@ def remove_col(self, col_name: str) -> "Schema":
     def without(self, col_names: List[str]) -> "Schema":
         return self.excluding_by_name(col_names)
 
+    def with_tags(self, col_names: List[str], tags: Union[str, Tags]):
```
oliverholworthy, line 441: Depending on what usage pattern is going to be more common. (changing more than one column with the same tags or more than one column with different tags).

We could consider combining the two args into a mapping from col name to tags `Dict[str, Union[str, Tags]]`? Same principle for `with_properties` / `with_dtypes` (the dtype one might be less natural to map to with the `is_list` `is_ragged` args?)

karlhigley: My thought here was that the most common use case is probably tagging a chunk of columns as categorical or continuous, in which case they'd all get the same new tags. Not sure if that's actually the most common case though.

karlhigley:
@oliverholworthy @nv-alaiacano Do y'all still want to move forward with this change? If so, what path would you foresee us taking to get the libraries updated to match?

PR #123, opened 2022-08-24
Check that `cudf is not None` before relying on its `Dataframe` type

karlhigley:


merlin/core/dispatch.py
```diff
@@ -75,7 +75,7 @@ def inner2(*args, **kwargs):
         return inner1
 
 
-if HAS_GPU:
+if HAS_GPU and cudf is not None:
```
oliverholworthy, line 78: If `cudf` failed to import above. It's wrapped in a try/except that should set HAS_GPU to False if it fails to import.

karlhigley: And yet somehow that isn't happening 😕

PR #125, opened 2022-08-25
Migrate Merlin DAG executors from NVTabular

karlhigley:
These executor classes were extracted from NVTabular's `Workflow` class, and provide general functionality for executing Merlin DAGs to be used in other Merlin libraries with DAGs (like Systems.)

Next steps after this PR:
* Update NVT [`Workflow`](https://github.com/NVIDIA-Merlin/NVTabular/blob/main/nvtabular/workflow/workflow.py#L75) and [integration tests](https://github.com/NVIDIA-Merlin/NVTabular/search?q=MerlinPythonExecutor) to use these executors and remove [the ones in NVT](https://github.com/NVIDIA-Merlin/NVTabular/blob/main/nvtabular/workflow/executor.py)
* Compare the `LocalExecutor` and `DaskExecutor` to how Merlin Models is [doing bulk model predictions](https://github.com/NVIDIA-Merlin/models/blob/4df954e0681ef55981fc93f723cbf110c716ab0b/merlin/models/tf/models/base.py#L797-L826)
* Write a `TritonPythonModel` that uses `LocalExecutor` to run the Systems operator graph
* Make an `AsyncLocalExecutor` as a drop-in replacement for `LocalExecutor` in the `TritonPythonModel` (to improve performance when IO-bound, like with network calls to a feature store)


karlhigley:
rerun tests

karlhigley:
rerun tests

oliverholworthy:
Are there any unit tests we can copy/adapt from NVTabular or are you thinking the tests in NVTabular will be sufficient now that we have those running alongside these (since #126) ?

karlhigley:
For now, I'm okay with relying on the downstream tests in NVTabular. Moving the code across doesn't make the test coverage any worse, but I agree it could stand to be improved substantially.

oliverholworthy (approved):

karlhigley:
rerun tests

jperez999 (approved):

jperez999:
seems like models is failing but I am not sure that has to do with this.

karlhigley:
Yeah, same, I think that's a pre-existing failure. I'm working on restructuring the way the downstream repo tests run to make several of these Core PRs mergeable.

PR #127, opened 2022-08-29
Split the downstream repo tests into separate Tox environments and Github actions

karlhigley:


karlhigley:
rerun tests

tox.ini
```diff
@@ -30,6 +30,63 @@ deps =
...
+sitepackages=true
+allowlist_externals = git
+deps =
+    -rrequirements.txt
+    -rrequirements-dev.txt
+commands =
+    ; the GIT_COMMIT env is the current commit of the core repo
+    ; NOTE!!!! We must clean this up in the jenkins configuration with `rm -rf "nvtabular-$GIT_COMMIT"`
+    git clone --depth 1 https://github.com/NVIDIA-Merlin/models.git models-{env:GIT_COMMIT}
+    git clone --depth 1 https://github.com/NVIDIA-Merlin/NVTabular.git nvtabular-{env:GIT_COMMIT}
```
oliverholworthy, line 63: do we need to install nvtabular from a git clone here? (vs. pip `python -m pip install --upgrade nvtabular@git+https://github.com/NVIDIA-Merlin/NVTabular.git` )

karlhigley: Yeah, we could do it that way too

PR #128, opened 2022-08-30
add subgraph feature of a Graph

nv-alaiacano:
This adds a `subgraph` field to a Graph. It is useful for defining and then selecting sections of the graph based on the provided output node. This will be used in NVTabular for selecting subsets of a Workflow that transforms User and Item features.

### DIFF
--- a/merlin/dag/graph.py
+++ b/merlin/dag/graph.py
@@ -16,6 +16,7 @@
 
 import logging
 from collections import deque
+from typing import Dict, Optional
 
 from merlin.dag.node import (
     Node,
@@ -30,8 +31,14 @@ LOG = logging.getLogger("merlin")
 
 
 class Graph:
-    def __init__(self, output_node: Node):
+    def __init__(self, output_node: Node, subgraphs: Optional[Dict[str, Node]] = None):
         self.output_node = output_node
+        self.subgraphs = subgraphs or {}
+
+    def subgraph(self, name: str):
+        if not name in self.subgraphs.keys():
+            raise ValueError(f"No subgraph named {name}. Options are: {self.subgraphs.keys()}")
+        return Graph(self.subgraphs[name])
 
     @property
     def input_dtypes(self):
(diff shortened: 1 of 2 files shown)


karlhigley (approved):

jperez999:
this is spectacular! @nv-alaiacano

nv-alaiacano:
I added one additional piece of logic, which is to ensure that the subgraph is part of the "main" graph. You can see from the test that this is a little weird.

```python
def test_subgraph_with_unrelated_subgraph():
    sg1 = ["a", "b"] >> BaseOperator()
    sg2 = ["a", "c"] >> BaseOperator()
    sg3 = ["c", "d"] >> BaseOperator()

    # same input as sg3, but not included in the combined Graph.
    unrelated = ["c", "d"] >> BaseOperator()

    combined = sg1 + sg2 + sg3

    with pytest.raises(ValueError):
        Graph(combined, subgraphs={"sub1": sg1, "unrelated": unrelated})
```

Even though `unrelated` and `sg3` are the same, we don't detect that `unrelated`'s output node is in the graph, and so this raises a ValueError. It could be rare to do something like this, but potentially confusing if someone does. What are your thoughts on it @karlhigley @jperez999 ?

merlin/dag/graph.py
```diff
@@ -30,8 +31,19 @@
...
 class Graph:
-    def __init__(self, output_node: Node):
+    def __init__(self, output_node: Node, subgraphs: Optional[Dict[str, Node]] = None):
         self.output_node = output_node
+        self.subgraphs = subgraphs or {}
+
+        parents_with_deps = self.output_node.parents_with_dependencies
+        for name, sg in self.subgraphs.items():
+            if sg not in parents_with_deps:
+                raise ValueError(f"Subgraph {name} does not exist in output_node.")
```
karlhigley, line 41: Maybe something like "The output node of subgraph {name} ({self.subgraphs[name]}) does not exist among the parents or dependencies of {output_node}"?

nv-alaiacano:
rerun tests

PR #129, opened 2022-09-01
Expand `models` testing in PR checks to include TF and other frameworks

karlhigley:


oliverholworthy (approved):

tox.ini
```diff
@@ -60,9 +60,8 @@ commands =
     ; the GIT_COMMIT env is the current commit of the core repo
     ; NOTE!!!! We must clean this up in the jenkins configuration with `rm -rf "nvtabular-$GIT_COMMIT"`
     git clone --depth 1 https://github.com/NVIDIA-Merlin/models.git models-{env:GIT_COMMIT}
-    git clone --depth 1 https://github.com/NVIDIA-Merlin/NVTabular.git nvtabular-{env:GIT_COMMIT}
-    python -m pip install --upgrade ./models-{env:GIT_COMMIT}
-    python -m pip install --upgrade ./nvtabular-{env:GIT_COMMIT}
+    python -m pip install --upgrade ./models-{env:GIT_COMMIT}[implicit,lightfm,tensorflow,xgboost]
```
oliverholworthy, line 63: I approved the PR, however looks like there's is at least one missing requirement to run the tests. Specified under the separate key `dev` (adding `dev` to this list here should get this to work). We could split out the test requirements from dev if we want to make [that list](https://github.com/NVIDIA-Merlin/models/blob/df0dca381918ee40c221d317f70022449c56d221/requirements/dev.txt) more minimal

karlhigley: I discovered that and tried to [separate the test and docs requirements](https://github.com/NVIDIA-Merlin/models/pull/708), but it looks like the Tensorflow tests still build the docs?

oliverholworthy: based on the structure of the repos (including this one). Looks to me like it should be covered by the `docs-sched-rebuild` and i imagine safe to remove those steps from the tensorflow workflow

PR #131, opened 2022-09-02
Add XGBoost to `merlin-models` CPU tests

karlhigley:


oliverholworthy (approved):

karlhigley:
@oliverholworthy Opened this as a separate PR since these tests seem to fail on the absence of `cupy`, but I'm not really sure why. `cupy` shouldn't be present in CPU-only environments, but something is still trying to use it anyway.

oliverholworthy:
> @oliverholworthy Opened this as a separate PR since these tests seem to fail on the absence of `cupy`, but I'm not really sure why. `cupy` shouldn't be present in CPU-only environments, but something is still trying to use it anyway.

Looks like the 'train traditional models' is using the `gpu_hist` objective which triggers a codepath in xgboost.dask to use cupy. we could either change the notebook (to use a different objective or pick the objective dynamically based on the GPU availability) or annotate the tests with a pytest skip if GPU is unavailable

oliverholworthy:
I think this is now unblocked. However we have two failing tests that are unrelated to this change:
- Models - `test_usecase_accelerate_training_by_lazyadam`
  I'm not sure what the issue here is exactly, possibly something about memory use.
- Systems - `test_ensemble`
  `AttributeError: 'SelectionOp' object has no attribute 'export_name'` being fixed in systems

karlhigley:
Fix for the systems issue in https://github.com/NVIDIA-Merlin/systems/pull/209. As long as we're confident this didn't cause the memory issue in the Models tests, probably fine to ignore the Systems test and merge this.

oliverholworthy:
I think we can be confident this didn't impact the models tests since we've seen the same error show up in more than one other PR now.

PR #135, opened 2022-09-07
add in asvdb for nvt testing in core

jperez999:


tox.ini
```diff
@@ -54,14 +54,17 @@ passenv=GIT_COMMIT
...
+    -r requirements-dev.txt
 commands =
     ; the GIT_COMMIT env is the current commit of the core repo
     ; NOTE!!!! We must clean this up in the jenkins configuration with `rm -rf "nvtabular-$GIT_COMMIT"`
     git clone --depth 1 https://github.com/NVIDIA-Merlin/models.git models-{env:GIT_COMMIT}
     python -m pip install --upgrade ./models-{env:GIT_COMMIT}[dev,implicit,lightfm,tensorflow]
-    python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/NVTabular.git
+    git clone --depth 1 https://github.com/NVIDIA-Merlin/NVTabular.git NVTabular-{env:GIT_COMMIT}
+    pip install --upgrade ./NVTabular-{env:GIT_COMMIT}
+    pip install -r .NVTabular-{env:GIT_COMMIT}/requirements/test.txt
```
karlhigley, line 66: We shouldn't be installing additional requirements files like this. They should either be listed in the `deps` section, or provided as a param on the install like `pip install nvtabular[test]`

karlhigley (changes requested):
We have cleaner ways to install additional requirements via `tox` deps or via install extras

tox.ini
jperez999: fixed

tox.ini
```diff
@@ -62,8 +62,7 @@ commands =
     git clone --depth 1 https://github.com/NVIDIA-Merlin/models.git models-{env:GIT_COMMIT}
     python -m pip install --upgrade ./models-{env:GIT_COMMIT}[dev,implicit,lightfm,tensorflow]
     git clone --depth 1 https://github.com/NVIDIA-Merlin/NVTabular.git NVTabular-{env:GIT_COMMIT}
-    pip install --upgrade ./NVTabular-{env:GIT_COMMIT}
-    pip install -r .NVTabular-{env:GIT_COMMIT}/requirements/test.txt
+    pip install --upgrade ./NVTabular-{env:GIT_COMMIT}[dev,test]
```
karlhigley, line 65: You're also gonna need `cpu` here

PR #139, opened 2022-09-13
Make it possible to use schema-awareness outside of operators and transform dictionary-like objects

karlhigley:
This should make it easier to use Merlin schemas outside the context of operators and DAGs without dataframes.

Included changes:
* Creates a `ComputeSchemaMixin` that contains only the schema computation parts of `BaseOperator`
* Creates `DictionaryLike`, `SeriesLike`, `DataframeLike`, and `Transformable` Python protocols
* Applies the new protocols throughout the `merlin.dag` and `merlin.dispatch` packages (replacing `DataFrameType` and `SeriesType`)
* Adds `DictArray` and `Column` classes that conform to the `DataFrameLike` and `SeriesLike` protocols (respectively)


merlin/core/dispatch.py
```diff
@@ -75,14 +76,6 @@ def inner2(*args, **kwargs):
         return inner1
 
 
-if HAS_GPU:
-    DataFrameType = Union[pd.DataFrame, cudf.DataFrame]  # type: ignore
-    SeriesType = Union[pd.Series, cudf.Series]  # type: ignore
-else:
-    DataFrameType = pd.DataFrame  # type: ignore
-    SeriesType = pd.Series  # type: ignore
-
```
karlhigley, line 84: These types have been a pain since they don't play nicely with Mypy, which doesn't like dynamically defined types like this. This PR replaces them with [Python protocols](https://peps.python.org/pep-0544/) that can be used as a static types and also checked at runtime (with `isinstance`.)

merlin/core/protocols.py
```diff
@@ -0,0 +1,116 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from typing import Protocol, runtime_checkable
+
+
+@runtime_checkable
+class DictLike(Protocol):
+    def __iter__(self):
+        return iter([])
```
karlhigley, line 22: The actual implementation of these methods don't matter, since they'll be overridden by anything that explicitly implements the protocol and otherwise are only used to check that matching method signatures are available on anything compared with `isinstance()`. These fake implementations make the linters happy though, since they insist that `__iter__()` must return an iterator and `__len__` must return a non-negative integer.

merlin/dag/executors.py
```diff
@@ -36,8 +39,35 @@ class LocalExecutor:
...
+        graphs = [Graph(deepcopy(nodes)[0]).construct_schema(schema) for schema in schemas]
+
+        results = []
+        for graph, data_ in zip(graphs, data):
+            results.append(self.transform(data_, [graph.output_node]))
+
+        return results
+
+    # TODO: Replace `nodes` with `graph` here?
     def transform(
```
karlhigley, line 64: This method has been long and hard to read for as long as it has existed (well before `Executor`s were a thing yet.) Here it's just broken down into some smaller methods that help explain what's happening via the method names to make it easier to read and think about.

merlin/dag/schema_mixin.py
```diff
@@ -0,0 +1,192 @@
...
+    CPU_DATAFRAME = auto()
+    # pandas dataframe
+    GPU_DATAFRAME = auto()
+    # dict of column name to numpy array
+    CPU_DICT_ARRAY = auto()
+    # dict of column name to cupy array
+    GPU_DICT_ARRAY = auto()
+
+
+class ComputeSchemaMixin:
```
karlhigley, line 37: This mixin allows non-operator classes to use the same schema computation machinery as operator without having to become operators and implement the full operator interface.

merlin/dag/selector.py
```diff
@@ -46,6 +46,12 @@ def __init__(
         self._tags = tags if tags is not None else []
         self.subgroups = subgroups if subgroups is not None else []
 
+        self.all = names == "*"
```
karlhigley, line 49: In order to make it possible to select all columns by default, which is useful in order to make the `ComputeSchemaMixin` methods have the lightest possible number of required parameters, we introduce the wildcard selector `"*"`, which is used in one of the following ways:
```
ColumnSelector("*")
"*" >> SomeOperator()
```

karlhigley: This is also useful for defining operator graphs that can be run on different input data by the same executor, since it allows late-binding to all provided columns in a similar way to how tags allow late-binding to some provided columns (but not others.)

pyproject.toml
```diff
@@ -32,7 +32,7 @@ ignore-nested-functions = true
 ignore-semiprivate = true
 ignore-setters = true
 fail-under = 70
-exclude = ["build", "docs", "merlin/io", "tests", "setup.py", "versioneer.py"]
+exclude = ["build", "docs", "merlin/core", "merlin/io", "tests", "setup.py", "versioneer.py"]
```
karlhigley, line 35: There's not really much reason to add docstrings to the newly defined protocols here, since classes that implement the protocols will have their own docstrings

tests/unit/dag/test_executors.py
```diff
@@ -0,0 +1,104 @@
...
+import numpy as np
+import pandas as pd
+
+from merlin.core.dispatch import make_df
+from merlin.dag import DictArray, Graph
+from merlin.dag.base_operator import BaseOperator
+from merlin.dag.executors import LocalExecutor
+from merlin.schema.schema import ColumnSchema, Schema
+
+
```
karlhigley, line 26: The tests in this file demonstrate the capabilities we discovered were needed in order to process the kinds of data we deal with after loading data from disk and before it gets passed to models, like operating on dictionary-like objects and transforming tuples of data (e.g. `(X, y)`) with the same transformation graph.

karlhigley:
The downstream tests are broken by the removal of `DataFrameType`, but it's straightforward to replace with the `DataFrameLike` protocol.

karlhigley:
@oliverholworthy @nv-alaiacano Thoughts on this? Downstream tests don't pass due to a breaking change related to `DataFrameType`, but this is ready for review when you get a chance.

merlin/core/dispatch.py
oliverholworthy: We could keep this code around for a follow-up PR to remove once the references to these in the other repos have been removed. That way we might be able to merge this without breaking tests in the other repos as a result, making the transition smoother?

merlin/dag/executors.py
```diff
@@ -46,93 +76,152 @@ def transform(
...
+    def _build_input_data(self, node, data, capture_dtypes=False):
+        """
+        Recurse through the graph executing parent and dependency operators
+        to form the input dataframe for each output node
+
+        Parameters
+        ----------
+        node : Node
+            Output node of the graph to execute
+        data : DataFrameType
```
oliverholworthy, line 104: input type here is now `Transformable`? and return type is the same or something else?

merlin/core/dispatch.py
karlhigley: Yeah, that's a good idea 👍🏻

merlin/core/protocols.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def __getitem__(self, index):
+        ...
+
+    def __eq__(self, other):
+        ...
+
+
+@runtime_checkable
+class Transformable(DictLike, Protocol):
```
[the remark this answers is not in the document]
karlhigley: I think we already use more than these three methods in practice (e.g. `__setitem__`), and part of what we're trying to do is make it possible to use dictionary-like objects, so even if we're not using the methods yet, this is laying the groundwork for being able to treat dictionaries and dataframes interchangeably.

karlhigley:
Breaking this down into smaller PRs, starting with:
* https://github.com/NVIDIA-Merlin/core/pull/140
* https://github.com/NVIDIA-Merlin/core/pull/141

karlhigley:
Superseded by the PRs listed above

PR #141, opened 2022-09-27
 Add and apply `DictArray` wrapper class and corresponding `Protocol` definitions

karlhigley:
These lay the foundation for being able to process dictionary-like objects interchangeably with dataframes in Merlin DAGs, which we expect to be useful for Systems (where we're already doing a simpler version of this), and potentially for models/dataloaders, which transform a lot of dictionary-like objects.

(Note that the wrapper class is called `DictArray` because that's the nomenclature that's [already used in other parts of the code](https://github.com/NVIDIA-Merlin/core/blob/3ea33c674fb3f9c8998a9674bbfbbbee3a387839/merlin/dag/base_operator.py#L33-L36) to describe this general type of data structure. I'm not attached to it, I just didn't want to create confusion by adding a second name and also didn't want to make changes to an `Enum` that's used in several libraries unless it was absolutely necessary.)


merlin/dag/dictarray.py
```diff
@@ -0,0 +1,121 @@
...
+# limitations under the License.
+#
+from typing import Dict, Optional
+
+import numpy as np
+
+from merlin.core.protocols import SeriesLike, Transformable
+
+
+class Column(SeriesLike):
```
karlhigley, line 23: In theory, this is where we might be able to put a shim in between the multiple arrays that represent a column and the code that processes them to make them look more like a single column. TBD what that would look like, but this creates a home for it.

merlin/core/dispatch.py
```diff
@@ -517,9 +518,9 @@ def detect_format(data):
             "csv": ExtData.CSV,
         }
         if isinstance(data, list) and data:
-            file_type = mapping.get(str(data[0]).split(".")[-1], None)
+            file_type = mapping.get(str(data[0]).rsplit(".", maxsplit=1)[-1], None)
         else:
-            file_type = mapping.get(str(data).split(".")[-1], None)
+            file_type = mapping.get(str(data).rsplit(".", maxsplit=1)[-1], None)
```
karlhigley, line 523: (These changes are required to appease the linters, but are otherwise unrelated to this PR)

merlin/core/dispatch.py
```diff
@@ -269,13 +270,13 @@ def series_has_nulls(s):
         return s.has_nulls
 
 
-def list_val_dtype(ser: SeriesType) -> np.dtype:
```
karlhigley, line 272: Replacing `SeriesType` and `DataframeType` (an ongoing pain with MyPy, which does not like dynamically defined types) with the protocols here doesn't functionally change much here, since the actual types don't prevent you from passing the wrong types into the function. The protocols (e.g. `SeriesLike`) can be checked at run-time via `isinstance(obj, protocol)` checks. Not sure if there's a way to enforce them automatically at runtime though.

merlin/core/protocols.py
```diff
@@ -0,0 +1,116 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from typing import Protocol, runtime_checkable
+
+
+@runtime_checkable
+class DictLike(Protocol):
```
karlhigley, line 20: This protocol definition contains the maximal set of methods that are shared by both actual dictionaries and dataframes (which are conceptually dictionaries of columns.)

merlin/core/protocols.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def __getitem__(self, index):
+        ...
+
+    def __eq__(self, other):
+        ...
+
+
+@runtime_checkable
+class Transformable(DictLike, Protocol):
```
karlhigley, line 68: In addition to the methods dictionary methods that are shared by dataframes, there are a few methods from dataframes that we use so frequently that it's easier to wrap a dictionary in a class and add them to the wrapper class than it would be to refactor the whole code base to do without them.

merlin/core/protocols.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def dtypes(self):
+        ...
+
+    def __getitem__(self, index):
+        ...
+
+
+@runtime_checkable
+class DataFrameLike(Transformable, Protocol):
```
karlhigley, line 81: This is the maximal set of methods shared by both Pandas dataframes and cuDF dataframes that aren't already part of the `Transformable` protocol. In theory, if there were another dataframe library that implemented the methods in this `Protocol` (e.g. Polars), we could use its dataframes in any place where we use the `DataFrameLike` type, but right now this protocol is only intended to match Pandas and cuDF dataframes.

merlin/dag/executors.py
```diff
@@ -48,110 +48,114 @@ def transform(
...
             else:
-                transformed_df = input_df
+                transformed_data = input_data
 
-            output_df = self._combine_node_outputs(node, transformed_df, output_df)
+            output_data = self._combine_node_outputs(node, transformed_data, output_data)
 
         if additional_columns:
-            output_df = concat_columns([output_df, df[_get_unique(additional_columns)]])
+            output_data = concat_columns(
```
oliverholworthy, line 66: It seems like if this line were ever reached it would throw an exception in the case of the non-dataframe input data

karlhigley: You’re right, it does! Have a fix for `concat_columns` that still needs to get pushed.

merlin/dag/executors.py
```diff
@@ -48,110 +48,114 @@ def transform(
...
+                    node, input_data, capture_dtypes=capture_dtypes
+                )
             else:
-                transformed_df = input_df
+                transformed_data = input_data
 
-            output_df = self._combine_node_outputs(node, transformed_df, output_df)
+            output_data = self._combine_node_outputs(node, transformed_data, output_data)
 
         if additional_columns:
```
oliverholworthy, line 65: this is a more general question that maybe doesn't affect this particular change, however it seems related enough to ask here: What is this `additional_columns` functionality for in the `LocalExecutor`? I can see it being used as part of the transform in the `DaskExecutor`. Would it simplify things if we dropped that here or is there a use-case not covered by the tests?

merlin/dag/executors.py
```diff
@@ -38,7 +38,7 @@ class LocalExecutor:
 
     def transform(
         self,
-        df,
+        transformable,
         nodes,
         output_dtypes=None,
```
oliverholworthy, line 43: similar to my question about `additional_columns` in that it's not strictly related to this change since it was here already and looks like mostly to match the interface of the `DaskExecutor`. However, it appears to be unused, wondering if it can be removed (in future change potentially). Or is it here as a placeholder to be used for something?

merlin/core/protocols.py
oliverholworthy: I'm wondering if the wrapper class is an implementation detail of the `LocalExecutor`, then it could potentially be encapsulated inside it so that regular dicts could be passed? Would that be a useful thing to do for the use-case in systems for example?

merlin/dag/executors.py
karlhigley: Added a commit to fix this

merlin/dag/executors.py
karlhigley: `additional_columns` is a way for nodes in the graph to fetch columns from the input data that are not transformed by the parent or dependency nodes in the graph. It's used in `DaskExecutor.fit()` when forming the transformed version of the data from previous nodes in order to compute statistics over it in order to merge in columns that don't need to be transformed by the upstream nodes. In order to do that, it gets passed into `DaskExecutor.transform()`, which then delegates the transformation of each Dask DataFrame partition to the `LocalExecutor.transform()`.

merlin/dag/executors.py
karlhigley: I think it's there to make the method signatures match across executors, since Dask uses it to give you partitions with the correct/user-specified dtypes in the output dataframe. How we want to handle arguments that some executors use and others don't was an open question, so I made the signatures match as a hopefully sensible default. This will probably become clearer if/when we try to add a third executor—possibly an async version of `LocalExecutor` or a distributed executor based on Ray.

merlin/core/protocols.py
karlhigley: It's possible that we could do that, but it seems like a potential instance of the [primitive obsession](https://wiki.c2.com/?PrimitiveObsession) code smell and would deprive us a place to add methods with additional functionality related to e.g. transferring data across frameworks or treating multiple arrays as a single column. 🤔

merlin/core/protocols.py
nv-alaiacano: Could you drop this paragraph and the one on `Transformable` into the docstring for the Protocol?

oliverholworthy (approved):

PR #142, opened 2022-09-27
Extract `ComputeSchemaMixin` from `BaseOperator`

karlhigley:
This potentially reduces the complexity of using Merlin schemas and the existing schema computation code outside the context of Merlin DAGs.

### DIFF
--- a/merlin/dag/schema_mixin.py
+++ b/merlin/dag/schema_mixin.py
@@ -211,6 +211,8 @@ class ComputeSchemaMixin:
         return col_schema.with_properties(properties)
 
     def _validate_matching_cols(self, schema: Schema, selector: ColumnSelector, method_name: str):
+        selector = selector or ColumnSelector()
+
         missing_cols = [name for name in selector.names if name not in schema.column_names]
         if missing_cols:
             raise ValueError(
--- a/merlin/dag/schema_mixin.py
+++ b/merlin/dag/schema_mixin.py
@@ -1,0 +1,224 @@
+#
+# Copyright (c) 2022, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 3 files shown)


merlin/dag/schema_mixin.py
```diff
@@ -0,0 +1,221 @@
...
+    GPU_DICT_ARRAY = auto()
+
+
+class ComputeSchemaMixin:
+    """
+    Adds the capability to compute input and output schemas to the classes it's mixed into
+    """
+
+    @property
+    def supports(self) -> Supports:
```
oliverholworthy, line 43: Where is this `supports` property used? 

I can see it being used in this `WorkflowRunner` in NVTabular.

https://github.com/NVIDIA-Merlin/NVTabular/blob/bc71785aab96368fb51f60d876fdfecba709e494/nvtabular/inference/workflow/base.py#L109-L112

If that's the only place and if we're planning on replacing that workflow with Merlin Systems then maybe we may soon not need this, or might come up with a new way of handling data format conversion between ops?

karlhigley: It’s used in the inference code, which swaps in optimized C++ implementations of some operators, depending on the value of the supports property. I think it will become more important rather than less as some of the other outstanding Core PRs related to processing dataframes and dictionary-like objects semi-interchangeably get merged, since not every operator will support both formats. (Many NVT operators will still only work dataframes, some Systems operators will likely be in at least a temporary state where they handle dict arrays but not dataframes.)

oliverholworthy: In what sense is this `supports` property relevant to this `ComputeSchemaMixin`? (If it's mostly about inference, I'm not seeing the connection to the schema propagation methods here.)

karlhigley: Anything that needs schema propagation does some kind of transform, which applies to certain kinds of data. In the case of Keras layers, that seems to be the “dict array” format. In order to mix-and-match objects that transform different types of data—which seems to be a consequence of the “we’ll just use any kind of object” approach—you have to know what kind of data it accepts.

I could see this living in `BaseOperator`, but at that point we’ve killed all hope of interoperability between different kinds of objects that implement transforms.

oliverholworthy (approved):
It's too early to say whether or not we will need this in practice. Can always revert if we need to since this doesn't change the way the operator works.

karlhigley:
It looks like "loader transforms" are already being added to Merlin Models, so it seems like we may not be far off from needing to figure it out. I'm inclined to either let this set open until we make a decision or close it and reopen later if we need to.

karlhigley:
@oliverholworthy I think I heard that some of the transforms that were potentially going to be run in the dataloaders have now been pulled into the models—is that accurate? If so, does that affect how we'd think about merging this PR (or not)?

PR #143, opened 2022-09-27
Add wildcard selector for cases where you'd like to select all columns

karlhigley:
The main use case for this is defining DAGs that can be applied to multiple dataframe-like/dictionary-like objects where the columns don't have overlapping tags (e.g. features vs labels, where the features are continuous and the labels are categorical.) There's not currently a way to define a DAG that's general enough to be applied that way, so this adds a more general selector type.

### DIFF
--- a/merlin/dag/selector.py
+++ b/merlin/dag/selector.py
@@ -46,6 +46,12 @@ class ColumnSelector:
         self._tags = tags if tags is not None else []
         self.subgroups = subgroups if subgroups is not None else []
 
+        self.all = names == "*"
+        if self.all:
+            self._names = []
+            self._tags = []
+            self._subgroups = []
+
         if isinstance(self._names, merlin.dag.Node):
             raise TypeError("ColumnSelectors can not contain Nodes")
 
@@ -103,7 +109,13 @@ class ColumnSelector:
             return self
         elif isinstance(other, merlin.dag.Node):
             return other + self
-        elif isinstance(other, ColumnSelector):
+
+        if self.all:
+            return self
+
+        if isinstance(other, ColumnSelector):
+            if other.all:
+                return other
 
             return ColumnSelector(
                 self._names + other._names,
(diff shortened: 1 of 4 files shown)


merlin/dag/selector.py
```diff
@@ -46,6 +46,12 @@ def __init__(
         self._tags = tags if tags is not None else []
         self.subgroups = subgroups if subgroups is not None else []
 
+        self.all = names == "*"
```
nv-alaiacano, line 49: names is annotated to be a `List[str]` and the docstring says it could also be a list of tuples of str. Should this check for those possible inputs as well?

merlin/dag/selector.py
```diff
@@ -130,14 +142,19 @@ def __rshift__(self, operator):
     def __eq__(self, other):
         if not isinstance(other, ColumnSelector):
             return False
-        return other._names == self._names and other.subgroups == self.subgroups
+        return other.all == self.all or (
```
nv-alaiacano, line 145: if `other.all` and `self.all` are both `False`, this will return `True`, ignoring the logic in the parentheses. 

I think this might want to be 
```python
    return (other.all and self.all) or (
        other._names == self._names and other.subgroups == self.subgroups
    )
```

merlin/dag/selector.py
oliverholworthy: supporting `["*"]` as a value that activates this functionality as well as `"*"`?

karlhigley: For consistency, we could make `["*"]` the syntax for it. That opens up the possibility of something like `["*", "column1"]` (redundant) or `[("*", "column1")]` (non-sensical), which we'd then have to handle.

karlhigley: (Tuples mean "group these columns together," which some operators use to understand how to apply their transforms. Categorify's "joint" and "combo" encoding modes are examples of cases where it's important to understand how columns should be grouped together.)

karlhigley: (For the record, I did not design this syntax or write the operators that rely on it, that was all before my time. ColumnSelector maintained it in the interest of avoiding breaking changes and, in all honesty, because no coherent alternative proposals have been put forward.)

nv-alaiacano (approved):

Issue #144, opened 2022-10-04
[Bug] Using numba to detect gpu availability breaks Dask-CUDA worker pinning


rjzamora:
While attempting to benchmark https://github.com/NVIDIA-Merlin/NVTabular/pull/1687, I discovered that the [dask-criteo benchmark](https://github.com/NVIDIA-Merlin/NVTabular/blob/main/bench/examples/dask-nvtabular-criteo-benchmark.py) does not work with the latest version of NVTabular/Merlin-core.

As far as I can tell, the problem is that https://github.com/NVIDIA-Merlin/core/pull/98 added the following logic to detect GPU availability: `HAS_GPU = len(cuda.gpus.lst) > 0`. This logic works just fine within a local process, but breaks Dask-CUDA device pinning when it is included in a top-level import (or is performed in the global context of the program). In other words, code like this **shouldn't** be executed by an import statement, like `from merlin.core.compat import HAS_GPU`.

The problem becomes apparent in a simple (Merlin-free) reproducer:

```python
# reproducer.py
from dask_cuda import LocalCUDACluster
from numba import cuda # This is fine

HAS_GPU = len(cuda.gpus.lst) > 0  # This is not fine

if __name__ == "__main__":
    cluster = LocalCUDACluster()
```

If you execute `python ./reproducer.py`, you sill see warnings like:

```
/.../distributed/distributed/comm/ucx.py:67: UserWarning: Worker with process ID 49507 should have a CUDA context assigned to device 1, but instead the CUDA context is on device 0. This is often the result of a CUDA-enabled library calling a CUDA runtime function before Dask-CUDA can spawn worker processes. Please make sure any such function calls don't happen at import time or in the global scope of a program.

PR #145, opened 2022-10-04
Avoid using numba to set device context in import

rjzamora:
Closes https://github.com/NVIDIA-Merlin/core/issues/144


merlin/core/compat.py
```diff
@@ -13,13 +13,22 @@
...
-    except cuda.cudadrv.error.CudaSupportError:
-        pass
 except ImportError:
     cuda = None
+
+HAS_GPU = False
+try:
+    import dask_cuda
+
+    HAS_GPU = dask_cuda.utils.get_gpu_count()
```
oliverholworthy, line 26: One edge case in CPU-mode where you might have `dask_cuda` installed but are not running with nvml support (For example, running a merlin docker image without GPU config). This call raises `pynvml.nvml.NVMLError_LibraryNotFound`. Might be able to add an extra `except` for that.

rjzamora: Thanks @oliverholworthy! I was about to ping you :)

Do you think it would be better to just try to import `pynvml` directly (rather than `dask_cuda`), and catch the `pynvml.nvml.NVMLError_LibraryNotFound` error?

merlin/core/compat.py
```diff
@@ -13,13 +13,22 @@
...
+except ImportError:
+    # Don't let numba.cuda set the context
+    # unless dask_cuda is not installed
+    if cuda is not None:
+        try:
+            HAS_GPU = len(cuda.gpus.lst) > 0
```
oliverholworthy, line 32: I made the change that added this check, and wasn't very familiar with the side-effects of `numba.gpus.lst` and it's relationship with `dask_cuda`. 

The motivation for the change was making the [automatic cluster type selection in the `Distributed` class](https://github.com/NVIDIA-Merlin/core/blob/v0.7.0/merlin/core/utils.py#L255) in `merlin.core.utils` work more reliably in contexts with and without GPU. Which was [previously using a successful import of `numba.cuda`](https://github.com/NVIDIA-Merlin/core/blob/v0.3.0/merlin/core/utils.py#L257) to determine whether or not to use the `dask_cuda.LocalCUDACluster`. 

Given that use-case, maybe we don't need anything in this except block? Since if we get to this point (without `dask_cuda`) and have visible cuda devices according to numba.cuda that doesn't help us anyway since we're using `dask_cuda` for the distributed functionality?

rjzamora: >I made the change that added this check, and wasn't very familiar with the side-effects of numba.gpus.lst and it's relationship with dask_cuda.

No worries - I wouldn't have had any idea that this was a problem if I hadn't been running the benchmark.

>Given that use-case, maybe we don't need anything in this except block?

Since `dask_cuda` depends on `pynvml` anyway, I suppose we can just simplify the `HAS_GPU` logic to be:

```python
try:
    import pynvml

    try:
        pynvml.nvmlInit()
        HAS_GPU = pynvml.nvmlDeviceGetCount() > 0
    except pynvml.nvml.NVMLError_LibraryNotFound:
        HAS_GPU = False
except ImportError:
    HAS_GPU = False
```

cc @pentschev (In case there is another *recommended* way to check for a cuda gpu on the system) :)

merlin/core/compat.py
oliverholworthy: Yes, I think using `pynvml` instead of `dask_cuda` here would work. And then maybe don't need the `numba.cuda.gpus.lst` check if we do that.

merlin/core/compat.py
oliverholworthy: I added this comment before seeing the suggestion about `pynvml` above. Doing that would keep the intent of this variable to indicate whether or not at least one GPU is visible. 

Assuming there's no side-effects of `pynvml.nvmlInit()` that would cause issues.

merlin/core/compat.py
```diff
@@ -17,9 +17,16 @@
 try:
     from numba import cuda
 
+except ImportError:
+    cuda = None
```
oliverholworthy, line 21: Might need to add the following line at the top of the file now. To avoid the linter raising errors about us not using cuda in here anymore

```
# pylint: disable=unused-import
```

merlin/core/compat.py
pentschev: Do you always have Distributed installed? If so, it may be simpler to just use https://github.com/dask/distributed/blob/e05c7a44b18e8f617681f5906054189bceb20f34/distributed/diagnostics/nvml.py#L113-L118 . The reason I suggest that is that for more complex cases (such as having PyNVML installed but no `libnvidia-ml.so` available, or no GPUs) there are other cases that may need handling, see https://github.com/dask/distributed/blob/e05c7a44b18e8f617681f5906054189bceb20f34/distributed/diagnostics/nvml.py#L59-L110 .

merlin/core/compat.py
rjzamora: Thanks! (forgot I didn't have pre-commit installed)

merlin/core/compat.py
oliverholworthy: [distributed is a dependency of this project](https://github.com/NVIDIA-Merlin/core/blob/v0.7.0/requirements.txt#L2) so,  we could use the `device_get_count` function from `distributed` if that simplifies handling the edge cases.

rjzamora: Okay, sounds good! We can let `distributed` worry about the status of nvml.

karlhigley (approved):

PR #152, opened 2022-10-13
necessary changes to allow graph execution in dataloader

jperez999:
This PR is necessary to allow the dataloader to be able to handle graph execution on a batch basis. Some of changes required here are not great this is because we dont have an  abstraction that is for merlin in general. We were trying to move towards this but after rigorous discussion it was decided those changes were unnecessary. The changes that would have made this much easier can be found here: https://github.com/NVIDIA-Merlin/systems/pull/159. Given that we deal with multiple different types of data all with their own dtypes to represent that data. As mentioned in Systems 159 a long time ago.

### DIFF
--- a/merlin/dag/executors.py
+++ b/merlin/dag/executors.py
@@ -120,7 +120,11 @@ class LocalExecutor:
         else:
             # If there are no parents, this is an input node,
             # so pull columns directly from root data
-            input_data = transformable[node_input_cols + list(addl_input_cols)]
+            if addl_input_cols:
+                addl_input_cols = list(addl_input_cols)
+            else:
+                addl_input_cols = []
+            input_data = transformable[node_input_cols + addl_input_cols]
 
         return input_data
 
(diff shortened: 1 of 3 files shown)


merlin/dag/executors.py
```diff
@@ -156,12 +160,17 @@ def _transform_data(self, node, input_data, capture_dtypes=False):
             # Update or validate output_data dtypes
             for col_name, output_col_schema in node.output_schema.column_schemas.items():
                 col_series = output_data[col_name]
-                col_dtype = col_series.dtype
+                col_dtype = col_series.dtype  # tf.int32
                 is_list = is_list_dtype(col_series)
 
                 if is_list:
                     col_dtype = list_val_dtype(col_series)
+                if hasattr(col_dtype, "as_numpy_dtype"):
```
jperez999, line 168: This is where having a "merlin" dtype would have been very helpful. We can avoid this code kludge.

karlhigley:
@EvenOldridge I think you were telling me that some of the transforms that would have run in the dataloaders are now part of the models, but there are still transforms to run at data-loading time (e.g. merging pre-trained embeddings into the examples.) This is one of the places where we're starting to hit issues with not having Merlin dtypes or cross-framework data transfer. How do you want to handle that? We can move forward with a pile of hacks to work around historically contentious design elements, or we can make decisions about how to build those elements (at least for now, knowing we might—and probably will—continue to iterate on them.)

oliverholworthy (approved):

PR #154, opened 2022-10-19
Fix dtype inference from pandas list column

rjzamora:
While working on https://github.com/NVIDIA-Merlin/NVTabular/pull/1693, I discovered that there is a subtle bug in `list_val_dtype` when the index of the target Series does not start with `0`. This PR modifies the pandas-specific logic to be a bit more general.


rjzamora:
Looks like CI failures are unrelated. I'm a bit confused by the error in the `df` fixture (Maybe the "id" column is somehow getting replaced by the "names" column in one of the tests?) -  No clue at all on the "Models Tests"

PR #159, opened 2022-11-01
Fix file-count warning in Dataset.to_parquet

rjzamora:
The `Dataset.to_parquet` API currently warns the user when the dataset does not contain enough partitions to produce the desired number of output files. However, this warning is currently **wrong** when the user specifies `out_files_per_proc>1`, because the current logic is only checking the number of subgraphs being used to write the output dataset (and each subgraph may be writing multiple files).

This PR adds a simple fix to check the actual number of files to be written before raising a warning.

### DIFF
--- a/merlin/io/dataset.py
+++ b/merlin/io/dataset.py
@@ -30,8 +30,13 @@ from dask.utils import natural_sort_key, parse_bytes
 from fsspec.core import get_fs_token_paths
 from fsspec.utils import stringify_path
 
-import merlin.core.dispatch as dispatch
-from merlin.core.dispatch import convert_data, hex_to_int, is_dataframe_object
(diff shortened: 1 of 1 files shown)


merlin/io/dataset.py
```diff
@@ -30,8 +30,13 @@
...
 
-import merlin.core.dispatch as dispatch
-from merlin.core.dispatch import convert_data, hex_to_int, is_dataframe_object
+from merlin.core.dispatch import (
+    convert_data,
+    hex_to_int,
+    is_dataframe_object,
+    is_list_dtype,
+    list_val_dtype,
+)
```
rjzamora, line 39: Just making the linter happy here.

karlhigley (approved):

PR #162, opened 2022-11-08
docs: Add preview doc builds for PRs

mikemckiernan:
Change the name of the upstream workflow
from `CPU CI` to `Core Tests (CPU)` so that
the `docs-preview-pr` workflow is triggered
after tests run successfully.


mikemckiernan:
I wanted to see the banner when I made #161 .  I assumed that you'd want this, but maybe you don't.

mikemckiernan:
Addressed in another PR.

PR #163, opened 2022-11-08
Remove explicit `DictArray` reference from `merlin.core.dispatch`

karlhigley:
This chunk of code makes it difficult to refactor Systems, since we'd like to copy `DictArray` into that code base while we work out what the new `Column` abstraction should look like with the plan to migrate the changes back to Merlin Core once we're happy with them and have the Systems tests passing.

The rewrite here changes the `isinstance()` checks to use protocols instead of the concrete `DictArray` class, which allows both the `DictArray` implementation here and the temporary one in Systems to be dispatched correctly.

### DIFF
--- a/merlin/core/dispatch.py
+++ b/merlin/core/dispatch.py
@@ -25,8 +25,7 @@ import pyarrow as pa
 import pyarrow.parquet as pq
 
 from merlin.core.compat import HAS_GPU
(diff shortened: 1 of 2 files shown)


karlhigley:
(The test failures in NVT and Models appear unrelated)

oliverholworthy (approved):

PR #164, opened 2022-11-09
Revert "Remove explicit `DictArray` reference from `merlin.core.dispatch`"

oliverholworthy:
Reverts NVIDIA-Merlin/core#163

This change resulted in errors running NVtabular workflows in other repos. Reverting to unblock other PRs while we investigate the reason for this.


oliverholworthy:
Revert not required. Found a solution that preservers the intent of the previous PR without the error #165

PR #165, opened 2022-11-10
Fix `concat_columns` for DataFrames with list features

oliverholworthy:
Follow-up to #163 which added an `isinstance` check for transformed data with the `DataFrameLike` protocol. That change resulted in an exception being rasiesd in NVTabular workflows that included both scalar and list features.

This PR:
-  Replaces the check against the protocol with the concrete classes (pandas/cudf)
- Adds a test for `concat_columns` with list and scalar features.

### Why does `isinstance(cudf_dataframe, DataFrameLike)` raise a `TypeError` ?

The reason for the type error comes from a combination of the  logic in the python `typing` module and a behaviour of `cudf` DataFrames.

A minimal example of what is going on in the `concat_columns` function is the following:

```python
import cudf
from typing import Protocol, runtime_checkable


@runtime_checkable
class ValuesAndColumns(Protocol):
    def values(self):
        ...

    @property
    def columns(self):
        ...

df = cudf.DataFrame({"a": [1], "b": [[2, 3]]})
isinstance(df, ValuesAndColumns)
# Raises TypeError: issubclass() arg 1 must be a class
```

There are two modifications that can be made to the above protocol that makes this instance check return without this exception.

- Remove `columns`
- Remove the `@property` on `columns`

The reason for this is that when a protocol is declared with an `@property`. The isinstance check does additional work that results in calling any properties that exist even if they were not declared as properties in the protocol.

In this example `values` is a property of the cudf.DataFrame. The instance check for the dataframe columns isn't a problem, but the side-effect of declaring it as a property in the protocol results in any other attributes in the protocol that happen to be properties on the instance being checked to be called.

Instead of returning from the first if block inside this instancecheck which  uses an `issubclass(instance.__class__, cls)`. We end up in the second if block which checks `getattr(instance, attr) is not None`

```python
class _ProtocolMeta(ABCMeta):
    # This metaclass is really unfortunate and exists only because of
    # the lack of __instancehook__.
    def __instancecheck__(cls, instance):
        # We need this method for situations where attributes are
        # assigned in __init__.
        if ((not getattr(cls, '_is_protocol', False) or
                _is_callable_members_only(cls)) and
                issubclass(instance.__class__, cls)):
            return True
        if cls._is_protocol:
            if all(hasattr(instance, attr) and
                    # All *methods* can be blocked by setting them to None.
                    (not callable(getattr(cls, attr, None)) or
                     getattr(instance, attr) is not None)
                    for attr in _get_protocol_attrs(cls)):
                return True
        return super().__instancecheck__(instance)
```
Source: https://github.com/python/cpython/blob/v3.8.15/Lib/typing.py#L1006-L1023

In the case of a cudf.DataFrame the `values` property is defined as:

```python
    @property
    def values(self):
        return self.to_cupy()
```

Source: https://github.com/rapidsai/cudf/blob/v22.10.01/python/cudf/cudf/core/frame.py#L416-L429

And there is [a bug in cudf](https://github.com/rapidsai/cudf/issues/12115) with a function `cudf.utils.dtypes.find_common_type`. If passed a scalar and list dtype, raises this TypeError exception.

Even if there was no bug in the `cudf` with list and scalar types, it is maybe not desirable to have an isinstance check peform a dataframe to cupy array conversion under the hood as part of that check. 

Some general principles with protocols we can learn from this:
- if we really need to check against them at runtime
  - avoid specifying protocol attributes with `@property` 
- Try to use checks against specific types where possible instead of a protocol

### DIFF
--- a/merlin/core/dispatch.py
+++ b/merlin/core/dispatch.py
@@ -27,7 +27,6 @@ import pyarrow.parquet as pq
 from merlin.core.compat import HAS_GPU
 from merlin.core.protocols import DataFrameLike, DictLike, SeriesLike
 
-
 cp = None
 cudf = None
 rmm = None
--- a/merlin/core/dispatch.py
+++ b/merlin/core/dispatch.py
@@ -25,7 +25,8 @@ import pyarrow as pa
 import pyarrow.parquet as pq
 
 from merlin.core.compat import HAS_GPU
-from merlin.core.protocols import DataFrameLike, DictLike, SeriesLike
+from merlin.core.protocols import DataFrameLike, SeriesLike
+
 
 cp = None
 cudf = None
@@ -350,23 +351,17 @@ def concat_columns(args: list):
     """Dispatch function to concatenate DataFrames with axis=1"""
     if len(args) == 1:
         return args[0]
-    elif isinstance(args[0], DataFrameLike):
+    elif isinstance(args[0], (cudf.DataFrame, pd.DataFrame)):
         _lib = cudf if HAS_GPU and isinstance(args[0], cudf.DataFrame) else pd
         return _lib.concat(
             [a.reset_index(drop=True) for a in args],
             axis=1,
         )
-    elif isinstance(args[0], DictLike):
+    else:
         result = type(args[0])()
         for arg in args:
             result.update(arg)
         return result
-    else:
-        _lib = cudf if HAS_GPU and isinstance(args[0], cudf.DataFrame) else pd
-        return _lib.concat(
-            [a.reset_index(drop=True) for a in args],
-            axis=1,
-        )
     return None
 
 
--- a/merlin/core/dispatch.py
+++ b/merlin/core/dispatch.py
@@ -25,7 +25,7 @@ import pyarrow as pa
 import pyarrow.parquet as pq
 
 from merlin.core.compat import HAS_GPU
-from merlin.core.protocols import DataFrameLike, SeriesLike
+from merlin.core.protocols import DataFrameLike, DictLike, SeriesLike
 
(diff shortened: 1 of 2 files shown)


karlhigley:
This is some really good detective work! I'm not sure that this fix will work either though, since we avoid checking against `cudf.DataFrame` because that won't be available in environments that don't have `cudf`.

> if we really need to check against them at runtime
Run-time checking is one of the main advantages of using them, since the alternative is to create explicit class hierarchies, which we can't do with types we don't own (e.g. `cudf.DataFrame`.)

> avoid specifying protocol attributes with @property
In order for any downstream class that implements the protocol to annotate itself as such and still satisfy `mypy`, its method declarations must match the way the methods are declared on the protocol. If the protocol doesn't declare a method as a property, the downstream class (e.g. `DictArray`) can't either, which affects the usage of the method in calling code: `obj.some_property` vs `obj.some_property()`. That has knock-on effects, because the intent of using a Protocol here is to create a category of objects that have interchangeable APIs including `cudf.DataFrame` and `DictArray.`
 
 > Try to use checks against specific types where possible instead of a protocol
This is often difficult due to the restriction that the checks must work in both GPU and CPU-only environments.

karlhigley:
One possibility would be to omit properties from the protocol definitions entirely. That would somewhat defeat the point of the `Transformable` protocol, but we'd at least still have the `dtypes` method as a way to differentiate between raw dictionaries and wrapped `DictArrays`.

oliverholworthy:
Updated to support CPU-only enviromnents by checking pandas and cudf separately

karlhigley (approved):

PR #170, opened 2022-11-20
Create standard Merlin dtypes in the `merlin.dtypes` module

karlhigley:
[Original Merlin Dtypes proposal slides](https://docs.google.com/presentation/d/1uo2FtTqGKTEt84VNHG7wbgW7ezNX8thfyFp1ZNxQ1YE/edit#slide=id.p)

This PR creates a `merlin.dtype` module that provides translations between a standard set of Merlin dtype descriptors and the dtype descriptors from external frameworks that we commonly use or interact with: Numpy, Pandas, cuDF, Tensorflow, Torch, and Triton. It also provides a mechanism for registering new sets of dtypes with mappings to Merlin dtypes. External frameworks are treated as optional and the relevant dtype mappings are only loaded/registered if the necessary packages are available (in order to avoid creating hard dependencies on e.g. deep learning frameworks or inference servers in Merlin Core.)

This has several benefits over using Numpy dtypes:
* We can directly represent unknown dtypes with a non-`None` value without defaulting them to `object` (like Numpy does)
* We can capture information in our dtypes that's used in Pandas/cuDF extension dtypes
* We're not forced to do lossy translations between external frameworks with Numpy as a constraining intermediary

It also lets us limit the set of dtype translation tables required to one, since we can use these Merlin dtypes as a Rosetta stone for translating between any pair of frameworks that each provide a Merlin dtype mapping.

There are related PRs adjusting downstream libraries to use these dtypes:
* NVTabular: https://github.com/NVIDIA-Merlin/NVTabular/pull/1707
* Models: https://github.com/NVIDIA-Merlin/models/pull/912
* Systems: https://github.com/NVIDIA-Merlin/systems/pull/239
* Dataloader: N/A (no changes required)

Some of the code this PR lays the groundwork for replacing:
* [Type normalization in `ColumnSchema`](https://github.com/NVIDIA-Merlin/core/blob/main/merlin/schema/schema.py#L74-L85)
* [Type translation when serializing `Schema` objects to Protobuf files](https://github.com/NVIDIA-Merlin/core/blob/main/merlin/schema/io/tensorflow_metadata.py#L31-L35)
* [Type translation between Numpy and DL frameworks in the dataloaders](https://github.com/NVIDIA-Merlin/dataloader/blob/015e85f76d2c139d924b12554e9793424a0419e2/merlin/loader/torch.py#L23-L35)
* [Type translation between Numpy and Triton types when writing Triton configs](https://github.com/NVIDIA-Merlin/systems/blob/1e2dc4803def8c37d1970ddd22631b1fb53fd91b/merlin/systems/triton/export.py#L709-L792)
* [Type translation between Numpy and Feast types when retrieving features](https://github.com/NVIDIA-Merlin/systems/blob/main/merlin/systems/dag/ops/feast.py#L12-L20)


jperez999 (approved):

merlin/dtype/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+    @property
+    def is_ragged(self):
+        return self.is_list and None in self.shape
+
+
+int32 = DType("int32", ElementType.Int, 32, signed=True)
+int64 = DType("int64", ElementType.Int, 64, signed=True)
+uint32 = DType("uint32", ElementType.UInt, 32)
+uint64 = DType("uint64", ElementType.UInt, 64)
+float32 = DType("float32", ElementType.Float, 32, signed=True)
```
angmc, line 59: Would we have to include integers and floats at lower bits like 16?

karlhigley: Yeah, that's actually required in order to make the NVT tests pass. The first commit on this branch was a rough sketch, but I'm fleshing it out as I test it with the other Merlin repos.

merlin/dag/executors.py
```diff
@@ -269,11 +270,15 @@ def transform(
         columns = list(flatten(wfn.output_columns.names for wfn in nodes))
         columns += additional_columns if additional_columns else []
 
+        if isinstance(output_dtypes, dict):
+            for col_name, col_dtype in output_dtypes.items():
+                output_dtypes[col_name] = col_dtype.to("numpy")
```
karlhigley, line 275: This can be `to_numpy` now, should be updated

merlin/schema/schema.py
```diff
@@ -263,11 +268,11 @@ def with_dtype(self, dtype, is_list: bool = None, is_ragged: bool = None) -> "Co
 
     @property
     def int_domain(self) -> Optional[Domain]:
-        return self._domain() if np.issubdtype(self.dtype, np.integer) else None
+        return self._domain() if self.dtype.elemtype.value == "int" else None
```
karlhigley, line 271: Might be nice to implement this logic as a convenience method on `DType` like `is_integer` 🤔

tests/unit/schema/test_column_schemas.py
```diff
@@ -13,15 +13,15 @@
...
 import pytest
 
+from merlin import dtype
 from merlin.schema import ColumnSchema
 from merlin.schema.schema import ColumnQuantity
 from merlin.schema.tags import Tags, TagSet
 
 
-@pytest.mark.parametrize("d_types", [numpy.float32, numpy.float64, numpy.uint32, numpy.uint64])
+@pytest.mark.parametrize("d_types", [dtype.float32, dtype.float64, dtype.uint32, dtype.uint64])
```
karlhigley, line 24: These dtypes should get normalized to Merlin types anyway when creating a `ColumnSchema`, so I'm not sure if this test actually needs to change. Should check that.

merlin/schema/io/tensorflow_metadata.py
```diff
@@ -181,6 +181,12 @@ def from_merlin_schema(cls, schema: MerlinSchema) -> "TensorflowMetadata":
         """
         features = []
         for col_name, col_schema in schema.column_schemas.items():
+            if col_schema.dtype is None:
+                raise ValueError(
+                    f"Schema for column `{col_name}` does not have a dtype. "
+                    "ColumnSchema objects must have dtypes in order to be serializable."
+                )
+
```
karlhigley, line 189: We do this validation here instead of the `ColumnSchema` constructor because it's sometimes convenient to allow column schemas to have blank dtypes (e.g. in the tests), but we can't serialize those column schemas to the Merlin schema format.

tests/unit/schema/test_column_schemas.py
karlhigley: The conversion still works fine with `numpy.whatever` types but that makes it harder to write the test assertions, so it actually does make sense to convert these over (just for convenience.)

tests/unit/schema/test_schema_io.py
```diff
@@ -32,6 +33,7 @@ def test_json_serialization_with_embedded_dicts():
         [
             ColumnSchema(
                 "userid",
+                dtype=dtype.int32,
```
karlhigley, line 36: These changes are due to the stricter validation of schemas for IO; we now need column schemas with dtypes specified. This used to work without, but that was somewhat accidental: the dtype `None` was translated by Numpy to `np.float64`, which then meant we had an incorrect dtype to write to the file, which was still enough for these tests to pass since they're not particular about the dtypes.

merlin/dag/executors.py
```diff
@@ -177,17 +178,27 @@ def _transform_data(self, node, input_data, capture_dtypes=False):
 
                 if is_list:
                     col_dtype = list_val_dtype(col_series)
-                if hasattr(col_dtype, "as_numpy_dtype"):
-                    col_dtype = col_dtype.as_numpy_dtype()
-                elif hasattr(col_series, "numpy"):
+
+                # TODO: Add a utility that condenses the known methods of fetching dtypes
+                # from series/arrays into a single function, so that Tensorflow specific
+                # code doesn't leak into the executors
```
karlhigley, line 184: The need for this new utility was already latent in the existing code, but it becomes a lot clearer once you have a cross-framework way to translate dtypes. Chose not to tackle it here since that would expand the scope of this already somewhat large PR.

karlhigley:
The tests on the downstream repos that run on this PR don't run against the downstream repo branches that adjust for these changes, so they'll continue failing here even if the PR is fine. The tests on the downstream repo PRs *do* run against this branch, so we should be looking at them to determine when we're ready to merge this.

merlin/dtypes/aliases.py
```diff
@@ -0,0 +1,52 @@
...
+datetime64s = DType("datetime64[s]", ElementType.DateTime, 64, ElementUnit.Second)
+datetime64ms = DType("datetime64[ms]", ElementType.DateTime, 64, ElementUnit.Millisecond)
+datetime64us = DType("datetime64[us]", ElementType.DateTime, 64, ElementUnit.Microsecond)
+datetime64ns = DType("datetime64[ns]", ElementType.DateTime, 64, ElementUnit.Nanosecond)
+
+# Miscellaneous
+string = DType("str", ElementType.String)
+boolean = DType("bool", ElementType.Bool)
+object_ = DType("object", ElementType.Object)
+unknown = DType("unknown", ElementType.Unknown)
```
jperez999, line 52: This dtype is essential. replacing none with our own logic, so we can catch.

jperez999:
Also didn't have a place to comment, but we may want to refactor the executor functionality to use the merlin dtypes to detect whether your a list or not. I dont know if that should happen by probing the dtype directly  (for is_list and is_ragged) or by refactoring is_list_dtype.

oliverholworthy (approved):

merlin/dtypes/base.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+@dataclass(eq=True, frozen=True)
+class DType:
+    """
+    Merlin dtypes are objects of this dataclass
+    """
+
+    name: str
+    elemtype: ElementType
```
oliverholworthy, line 68: one small question about naming. If users of this DType class are expected to use the element type, might be clearer here to use the full name for this `element_type` and same for the below two attrs?

karlhigley: Yeah, that's a reasonable suggestion. These started out following the `itemtype` convention from `numpy`, which is how this ended up as `elemtype`. There's something to be said for brevity, but I agree it's clearer to use the full words for the attribute names.

Issue #176, opened 2022-11-24
dask read_parquet changes disable parquet file statistics reading in default case


eordentlich:
Changes introduced in this dask PR https://github.com/dask/dask/pull/8992, especially [here](https://github.com/dask/dask/commit/3715ab5d68f944ef8bdcd9600e74aaa6c7ebf2ae#diff-ebd96924232465e8a6c3f41074e2cdfeb7e09f70de6eca1567be913762d3178fR362) and [here](https://github.com/dask/dask/commit/3715ab5d68f944ef8bdcd9600e74aaa6c7ebf2ae#diff-ebd96924232465e8a6c3f41074e2cdfeb7e09f70de6eca1567be913762d3178fR434) break/change the [default case logic](https://github.com/NVIDIA-Merlin/core/blob/v0.9.0/merlin/io/parquet.py#L217-L218) in nvtabular for reading parquet file statistics (to effectively disable it).  

This introduced a significant slow down (10-15 min) at start up (and possibly other problems) in our use case with KerasSequenceLoader and the large criteo dataset, as the entire dataset was being read in to determine the number of rows.

Work around is to pin dask version or explicitly set `gather_statistics=True` (or `calculate_divisions=True` to avoid dask deprecation warning) as argument to `Dataset` constructor, but would be good to have default behavior remain as before.

karlhigley:
Thoughts on how to address this, @rjzamora? The internals of `merlin.io.Dataset` are outside my wheelhouse 😅

rjzamora:
Thanks for raising this @eordentlich (and sorry for letting this regression slip by).  `Dataset` is in need of a general update. I should be able to work on it this week (and optimized `len` after `read_parquet` will be a top priority).

PR #178, opened 2022-11-30
Read parquet statistics to optimize len when they are missing

rjzamora:
Closes https://github.com/NVIDIA-Merlin/core/issues/176

Since `dd.read_parquet` will usually avoid reading parquet statistics, there are many cases were Merlin cannot use the dask-provided statistics to optimize operations like `len(ds)`. This PR adds some simple logic to update the missing statistics when needed.


karlhigley (approved):
Pretty sure the NVT test failures are unrelated since I just merged another PR intended to fix them

PR #182, opened 2022-12-14
Save schema with consistent dtype when `dtypes` is used

oliverholworthy:
When `Dataset.to_parquet` is used with `dtypes`, ensure we save the schema with a dtype that matches the parquet file dtypes.

## Motivation

Noticed this issue from the notebooks in Merlin when we updated to use the merlin-dataloader package which returns data with dtypes that match the parquet file.

[Fixed this with careful dtype matching here](https://github.com/NVIDIA-Merlin/Merlin/pull/765/commits/ab6321f38bdde4e3f84b09baadec73ab5143e331). However, this dtype PR would have made that commit unnecessary.


karlhigley (dismissed):

karlhigley:
@oliverholworthy It looks like this change may cause some dtype issues in NVTabular

merlin/core/dispatch.py
```diff
@@ -414,8 +414,10 @@ def parquet_writer_dispatch(df: DataFrameLike, path=None, **kwargs):
             _args.append(pa.Table.from_pandas(df, preserve_index=False).schema)
     elif cudf is not None:
         _cls = cudf.io.parquet.ParquetWriter
-    else:
-        ValueError("Unable to load cudf. Please check your environment GPU and cudf available.")
+    elif cudf is None:
+        ValueError(
```
karlhigley, line 418: I think this should raise?

merlin/io/dataset.py
```diff
@@ -892,14 +893,18 @@ def to_parquet(
                         output_files[fn + suffix] = rgs
             suffix = ""  # Don't add a suffix later - Names already include it
 
+        schema = deepcopy(self.schema)
```
oliverholworthy, line 896: I tried `Schema(self.schema.column_schemas)` here, but that' didn't seem to create a sufficiently distinct copy. The dtype changes further down were changing the `self.schema` dtypes as well as the schema copy dtypes

merlin/core/dispatch.py
oliverholworthy: 🤦 I was looking at this and wondering why pylint was confused about not reaching this line. And somehow I was blind to the missing raise here!

merlin/io/dataset.py
karlhigley: I think the reason is probably that `self.schema.column_schemas` is a dictionary, so `Schema(self.schema.column_schemas)` ends up using the same dictionary in both places.

karlhigley (approved):

PR #190, opened 2022-12-16
Add an ADR for creating Merlin dtypes

karlhigley:


mikemckiernan:
No feedback of consequence, but a sugg to use one sentence on each line.  It makes review a little easier.

karlhigley:
@mikemckiernan I'm not really sure where/how we're going to record these decisions, since it seems like the team wants to use Google Docs. I'm going to close these PRs for now, and we'll figure out where to store adopted proposals once we have one, I guess.

PR #195, opened 2023-01-06
 Add a new `shape` field to `ColumnSchema`

karlhigley:
This creates a place to store shape information for all dimensions of the data across both array/tensor and dataframe formats. In contrast to the existing "value_count" property (which only records the value counts of the lists in list field, this attribute is intended to capture the size of _all_ dimensions of the data (the batch dimension, the list lengths, embedding sizes, etc.)

A few significant design elements:
* Defining a `Shape` dataclass allows us to make them immutable, which we wouldn't get if we sub-classed `tuple` like some other frameworks do. For compatibility with those other frameworks (and just general ease of use), the `Shape` constructor accepts a tuple of dimensions, so it should be relatively straightforward to (for example) create a Merlin shape from a Tensorflow shape.
* Shapes are stored in dtypes, since shape information is how we know a particular column is a list, and several frameworks we use (e.g. cuDF, Feast) have either a general `List` dtype or specific list dtypes like `int32_list` for each element type. The API has been designed to hide this from calling code as much as possible, but it does influence how shapes work to some extent.
* The `is_list` and `is_ragged` flags are now computed from the shape and can no longer be set by providing them as constructor args. We tried to find a way to maintain that, but it's not possible to use dataclass init vars with the same name as a defined property. The best workaround for this we could come up with was to allow `is_list` and `is_ragged` to be set from the `properties` dictionary, in which case we do our best to infer a shape from the values of these flags.

### DIFF
--- a/merlin/dtypes/base.py
+++ b/merlin/dtypes/base.py
@@ -109,7 +109,7 @@ class DType:
             ) from exc
 
         try:
-            return mapping.from_merlin(self)
+            return mapping.from_merlin(replace(self, shape=None))
         except KeyError as exc:
             raise ValueError(
                 f"The registered dtype mapping for {mapping_name} doesn't contain type {self.name}."
--- a/merlin/dtypes/base.py
+++ b/merlin/dtypes/base.py
@@ -72,6 +72,10 @@ class DType:
     signed: Optional[bool] = None
     shape: Optional[Shape] = None
 
+    def __post_init__(self):
+        if not self.shape:
+            object.__setattr__(self, "shape", Shape())
+
     def to(self, mapping_name: str):
         """
         Convert this Merlin dtype to another framework's dtypes
--- a/merlin/dtypes/base.py
+++ b/merlin/dtypes/base.py
@@ -161,3 +161,15 @@ class DType:
             )
 
         return replace(self, shape=shape)
+
+    @property
+    def without_shape(self):
+        """
+        Create a copy of this object without the shape
+
+        Returns
+        -------
+        DType
+            A copy of this object with the shape removed
+        """
+        return self.with_shape(Shape())
--- a/merlin/dtypes/base.py
+++ b/merlin/dtypes/base.py
@@ -14,11 +14,12 @@
 # limitations under the License.
 #
 
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from enum import Enum
-from typing import Optional
+from typing import Optional, Tuple, Union
 
 from merlin.dtypes.registry import _dtype_registry
+from merlin.dtypes.shape import Shape
 
 
 class ElementType(Enum):
@@ -69,6 +70,7 @@ class DType:
     element_size: Optional[int] = None
     element_unit: Optional[ElementUnit] = None
     signed: Optional[bool] = None
+    shape: Optional[Shape] = None
 
     def to(self, mapping_name: str):
         """
@@ -125,3 +127,33 @@ class DType:
     @property
     def is_float(self):
         return self.element_type.value == "float"
+
+    def with_shape(self, shape: Union[Tuple, Shape]):
+        """
+        Create a copy of this dtype with a new shape
+
+        Parameters
+        ----------
+        shape : Union[Tuple, Shape]
+            Object to set as shape of dtype, must be either a tuple or Shape.
+
+        Returns
+        -------
+        DType
+            A copy of this dtype containing the provided shape value
+
+        Raises
+        ------
+        TypeError
+            If value is not either a tuple or a Shape
+        """
+        if isinstance(shape, tuple):
+            shape = Shape(shape)
+
+        if not isinstance(shape, Shape):
+            raise TypeError(
+                f"Provided value {shape} (of type {type(shape)}) for DType.shape property "
+                "is not of type Shape."
+            )
+
+        return replace(self, shape=shape)
--- a/merlin/dtypes/base.py
+++ b/merlin/dtypes/base.py
@@ -109,7 +109,7 @@ class DType:
             ) from exc
 
         try:
-            return mapping.from_merlin(replace(self, shape=None))
+            return mapping.from_merlin(self.without_shape)
         except KeyError as exc:
             raise ValueError(
                 f"The registered dtype mapping for {mapping_name} doesn't contain type {self.name}."
--- a/merlin/dtypes/shape.py
+++ b/merlin/dtypes/shape.py
@@ -78,10 +78,12 @@ class Shape:
                     new_dim = Dimension(dim[0], dim[1])
                 elif isinstance(dim, int):
                     new_dim = Dimension(dim, dim)
+                elif dim is None:
+                    new_dim = Dimension()
                 else:
                     raise ValueError(
                         f"Invalid shape tuple format: {self.dims}. Each dimension is expected "
-                        " to be either a single integer or a length 2 tuple."
+                        " to be None, a single integer, or a tuple with length 2."
                     )
                 new_dims.append(new_dim)
 
--- a/merlin/dtypes/shape.py
+++ b/merlin/dtypes/shape.py
@@ -111,10 +111,6 @@ class Shape:
         return tuple(dim.max for dim in self.dims)
 
     @property
-    def fixed(self) -> Tuple:
-        return tuple(dim.min if dim.is_fixed else None for dim in self.dims)
-
-    @property
     def is_bounded(self):
         return all(dim.is_bounded for dim in self.dims)
 
--- a/merlin/dtypes/shape.py
+++ b/merlin/dtypes/shape.py
@@ -1,0 +1,119 @@
+#
+# Copyright (c) 2022, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from dataclasses import dataclass, field
+from typing import Optional, Tuple
+
+
+@dataclass(frozen=True)
+class Dimension:
+    """
+    The range of potential sizes for a single dimension of a field or column
+    """
+
+    min: int = 0
+    max: Optional[int] = None
+
+    def __post_init__(self):
+        if self.min is None:
+            raise ValueError("The minimum size of a dimension cannot be None. ")
+
+        if self.min < 0:
+            raise ValueError(
+                "The minimum size of a dimension must be non-negative. " f"Provided min: {self.min}"
+            )
+
+        if self.max and self.max < 0:
+            raise ValueError(
+                "The maximum size of a dimension must be at least one. " f"Provided max: {self.max}"
+            )
+
+        if self.max and self.max < self.min:
+            raise ValueError(
+                "The maximum size of a dimension must be at least as large as the minimum size. "
+                f"Provided min: {self.min} max: {self.max}"
+            )
+
+    @property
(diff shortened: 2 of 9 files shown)


merlin/schema/schema.py
```diff
@@ -240,12 +264,14 @@ def value_count(self) -> Optional[Domain]:
...
-            other.dtype, is_list=other.is_list, is_ragged=other.is_ragged
+        col_schema = (
+            self.with_name(other.name)
+            .with_dtype(other.dtype)
+            .with_tags(other.tags)
+            .with_properties(other.properties)
+            .with_properties({"is_list": other.is_list, "is_ragged": other.is_ragged})
         )
-        col_schema = col_schema.with_name(other.name)
+        # TODO: Figure out if and how to merge shapes here too
```
EvenOldridge, line 274: Is this deliberately left to future work or do we need a way to merge shapes still?

EvenOldridge (approved):
I like the level of detail into the error messaging and the comprehensive tests.  Assuming the merge of shapes is meant to be future work this looks good.

merlin/schema/schema.py
karlhigley: Yeah, this should get filled in. The tricky part is figuring out what it means to merge shapes. 😅

Since you can have unbounded dimensions with `min=0, max=None`, one possible way to define it would be to take the shape from each column schema and:
- Check that the shapes have the same number of dimensions, if not raise an error
- For each dimension:
  - Check if the two versions of the dim are identical, if so use either one
  - Check if one is bounded and the other is unbounded, if so use the bounded version
  - Otherwise, we have two bounded versions of the dim that don't match, and we can either:
    - Choose whether the left or right side of the merge gets priority and overwrites the other
    - Raise an error

merlin/dtypes/base.py
```diff
@@ -103,7 +109,7 @@ def to(self, mapping_name: str):
             ) from exc
 
         try:
-            return mapping.from_merlin(self)
+            return mapping.from_merlin(replace(self, shape=None))
```
karlhigley, line 112: This should use the `.without_shape` method defined below

merlin/dtypes/shape.py
```diff
@@ -0,0 +1,135 @@
...
+            for i, dim in enumerate(self.dims):
+                if isinstance(dim, Dimension):
+                    new_dim = dim
+                elif isinstance(dim, tuple) and len(dim) == 2:
+                    new_dim = Dimension(dim[0], dim[1])
+                elif isinstance(dim, int):
+                    new_dim = Dimension(dim, dim)
+                else:
+                    raise ValueError(
+                        "Invalid shape tuple format: {self.dims}. Each dimension is expected to be "
```
karlhigley, line 83: This should be an f-string

merlin/dtypes/shape.py
```diff
@@ -0,0 +1,135 @@
...
+    def min(self) -> Tuple:
+        return tuple(dim.min for dim in self.dims)
+
+    @property
+    def max(self) -> Tuple:
+        return tuple(dim.max for dim in self.dims)
+
+    @property
+    def fixed(self) -> Tuple:
+        return tuple(dim.min if dim.is_fixed else None for dim in self.dims)
```
karlhigley, line 115: This method returns the shape with only the fixed dimensions included and the rest marked with `None` as a placeholder. I'm a bit worried that the distinction between `fixed` and `is_fixed` isn't clear, but I'm not sure what else to call them to make it clearer.

merlin/schema/schema.py
```diff
@@ -82,37 +82,47 @@ def __post_init__(self):
...
+    def shape(self):
+        return self.dtype.shape
+
+    @property
+    def is_list(self):
+        return self.dtype.shape.is_list
+
+    @property
+    def is_ragged(self):
+        return self.dtype.shape.is_ragged
```
karlhigley, line 125: We might want to add `is_fixed`/`is_variable` here too, so that you don't have to reach through the dtype and the shape to access them (`col_schema.dtype.shape.is_variable`)

bschifferer:
I reviewed the code. It seems good to me/makes sense - however, I haven't worked in that deep level of Merlin/core to confident approve it

merlin/dtypes/shape.py
oliverholworthy: what's the use-case for wanting to return this tuple of fixed dims sizes?

tests/unit/dtypes/test_shape.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def test_min_max_val_dimension():
+    dim = Dimension(2, 3)
+    assert dim.min == 2
+    assert dim.max == 3
+
+
+def test_fixed_min_with_unbounded_max():
+    dim = Dimension(2)
```
oliverholworthy, line 37: It may make our API more familiar to new users to make `Dimension(2)` correspond to a fixed dimension of size 2 (min=2, max=2) instead of a ragged dimension with min size of 2

oliverholworthy: I haven't seen any mentions of use-cases motivating the addition of the `min` in the dimension size other than it being there before in the value count property. I guess it's a more generic version of a Dimension in tensorflow for example that is either a fixed size or ragged with unknown min/max value.

oliverholworthy: We may need to give some consideration to what the min/max means in the ragged case. For example the ValueCount operator in NVTabular would result in a schema  with min and max set to the min/max observed in the data it was fit on. However, that doesn't necessarily imply that the model or serving system is limited to those min/max values

karlhigley: > It may make our API more familiar to new users to make Dimension(2) correspond to a fixed dimension of size 2 (min=2, max=2) instead of a ragged dimension with min size of 2

We considered this, but it made the code rather difficult to write, so we settled on providing syntactic sugar when creating a `ColumnSchema` with the `dims` property. We expect people to either use that path or the `Shape` constructor, which offers the same functionality, instead of interacting with the Dimension constructor directly. In either of those places, you can supply `2` for a dim, which means a dimension of fixed size like you're envisioning.

karlhigley: > I haven't seen any mentions of use-cases motivating the addition of the min in the dimension size other than it being there before in the value count property. I guess it's a more generic version of a Dimension in tensorflow for example that is either a fixed size or ragged with unknown min/max value.

We're not quite there yet, but we should be able to directly validate a schema against a `DictArray`/`TensorTable` and match up the shapes of the data with the requirements in the schema. Allowing a non-zero minimum lets us differentiate between list fields that are allowed to be empty and those that aren't. I'm not sure how useful that will end up being though.

karlhigley: > We may need to give some consideration to what the min/max means in the ragged case. For example the ValueCount operator in NVTabular would result in a schema with min and max set to the min/max observed in the data it was fit on. However, that doesn't necessarily imply that the model or serving system is limited to those min/max values

Yeah, this is a good point. What I'm envisioning is that each Merlin component updates the schema and stores the updated version alongside its artifacts, so when we save a model if the model is fine with a column having a minimum size of zero, it should update the schema to say so.

Something like this:
```
Parquet files -> NVT inferred schema -> NVT DAG -> NVT annotated schema -> Parquet files w/ annotated schema

Annotated schema -> dataloader transforms -> dataloader output schema -> model input layer

*model training happens*

Saved model w/ input and output schemas -> Systems op
```

merlin/schema/schema.py
```diff
@@ -82,37 +82,47 @@ def __post_init__(self):
...
+
+        if new_shape:
+            object.__setattr__(self, "dtype", self.dtype.with_shape(new_shape))
+
+    @property
+    def shape(self):
+        return self.dtype.shape
+
+    @property
+    def is_list(self):
```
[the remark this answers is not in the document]
karlhigley: > I'm wondering if we can avoid the initial breaking change from this PR by keeping the is_list, is_ragged as properties that get set appropriately in the post_init and update methods based on the shape?

Yeah, we could try it.

merlin/dtypes/shape.py
karlhigley: Let's remove it and see if we need to add it at some point. Seemed like a handy utility for handling things like `(64, (0, None), 128)` where you have a batch from the dataloader containing embedding lists of indeterminate length, but maybe I've overthought that part.

merlin/schema/schema.py
karlhigley: For now, I'm just going to give the same priority to the shape from `other` as the rest of the attributes. We'll see if that causes problems and adjust if necessary.

karlhigley:
@oliverholworthy I updated this PR to keep `is_list` and `is_ragged` as properties that are set in the `__post_init__` to match the shape, which really does clean things up quite a bit. I think it still ends up being a breaking change that requires coordination with the other repos though, because validating that the provided flags are consistent with the shape turns up places where that isn't currently true.

tests/unit/dtypes/test_shape.py
oliverholworthy: > We considered this, but it made the code rather difficult to write, so we settled on providing syntactic sugar when creating a `ColumnSchema` with the `dims` property. [...] you can supply `2` for a dim, which means a dimension of fixed size like you're envisioning.

I'd expect to be able to proivde `None` too as part of this syntactic sugar. For example, `Shape((1, None, 10))` to create a shape with 2 fixed and one ragged dimension. This currently raises a ValueError.

oliverholworthy: Supporting `None` as a dimension to the Shape constructor would help with making it _relatively straightforward to (for example) create a Merlin shape from a Tensorflow shape._

oliverholworthy:
> I think it still ends up being a breaking change that requires coordination with the other repos though, because validating that the provided flags are consistent with the shape turns up places where that isn't currently true.

Looks like we'll be setting a requirement for the shape to be specified if `is_ragged=False` is specified, the fixed dimension will be required to be passed. Would it be possible or helpful with the migration to loosen that restriction further and only check the shape if it's used explictly (either with the `dims` passed to the `ColumnSchema` constructor or `with_shape`)

karlhigley:
> Would it be possible or helpful with the migration to loosen that restriction further and only check the shape if it's used explictly?

I shifted the validation in this general direction, so that only information that's explicitly provided gets validated. I'm not sure about lifting the restriction that some kind of shape info has to be provided (through the dtype's shape, the dims, or the value counts) when `is_ragged=False` because that will put us in a position where we sometimes don't have a shape. That may complicate the downstream changes as we try to migrate over. 🤔

oliverholworthy:
> because that will put us in a position where we sometimes don't have a shape. That may complicate the downstream changes as we try to migrate over

I suppose we're already in the position where we sometimes don't have a shape, since there's nothing in here currently to enforce having a value count when `is_ragged=False`. Adding this requirement in this PR forces us to address that now. I support that choice if we're commited to making the changes to the downstream libs right now. If it's mostly a question of how we'll identify which parts of the code need to change, we could achieve that with an optional strict mode or something along those lines.

And closely related to this: to what extent will this change make to loading previously saved schemas (e.g. from an NVTabular workflow and saved schema file). Will we be able to load schemas saved with a prior version after merging this? We'd presumably prefer to be backwards compatible if possible, however if it's a necessary breaking change across the saved file format, then we might need some extra docs about this when we release the next version.

karlhigley:
> Adding this requirement in this PR forces us to address that now. I support that choice if we're commited to making the changes to the downstream libs right now. If it's mostly a question of how we'll identify which parts of the code need to change, we could achieve that with an optional strict mode or something along those lines.

We're thinking along similar lines then; we've disabled the `is_ragged=False` without shape info validation for now, and will use it to identify the places where changes are needed in further dev work. 👍🏻 

> to what extent will this change make to loading previously saved schemas (e.g. from an NVTabular workflow and saved schema file). Will we be able to load schemas saved with a prior version after merging this? We'd presumably prefer to be backwards compatible if possible, however if it's a necessary breaking change across the saved file format, then we might need some extra docs about this when we release the next version.

I believe the change is backward but maybe not forward compatible, since the allowed values of value counts are slightly different in a more permissive direction after this change.

oliverholworthy (approved):
Looks like this is in good shape! ✨

PR #201, opened 2023-01-25
Bump `dask`, `distributed`, `fsspec` versions to `>=2022.7.0`

karlhigley:


jperez999 (approved):

karlhigley:
This version bump doesn't actually seem to work 😅

karlhigley:
@jperez999 Seeing a bunch of errors like this in the Systems tests:
 ```
AttributeError: 'Workflow' object has no attribute '_transform_df'
```
Seems like maybe there's a chunk of code for either Core or NVT somewhere that didn't pushed. Any idea if you have something stashed somewhere?

File: requirements.txt [hunk 1]
```diff
@@ -1,11 +1,11 @@
-dask>=2022.3.0
-distributed>=2022.3.0
+dask==2022.7.1
```
oliverholworthy, line 1: Since this is going to turn into the package `install_requires`. having a range (minimum) may be preferable to the explicit pinned version here to improve the chances of finding compatibile versions of packages in a project?

karlhigley: I had trouble with that because it ended up picking newer versions than we're actually compatible with, but I'm not sure where the cut-off is.

PR #202, opened 2023-01-25
dict array migration to core GPU and CPU conversion foundation

jperez999:
This provides the foundation for the array conversion capability. It will be the basis for allowing data to transfer from GPU to CPU and between frameworks.


karlhigley:
Superseded by `TensorTable`

PR #203, opened 2023-01-31
Use bitwise or operator to merge `ColumnSchema`s

jperez999:
This makes merging two column schemas easier by using the bitwise OR operator ( | ).


merlin/schema/schema.py
```diff
@@ -239,13 +239,15 @@ def value_count(self) -> Optional[Domain]:
         value_count = self.properties.get("value_count")
         return Domain(**value_count) if value_count else None
 
-    def __merge__(self, other):
-        col_schema = self.with_tags(other.tags)
-        col_schema = col_schema.with_properties(other.properties)
-        col_schema = col_schema.with_dtype(
-            other.dtype, is_list=other.is_list, is_ragged=other.is_ragged
+    def __or__(self, other):
+        dtype = self.dtype if self.dtype != md.unknown else other.dtype
```
jperez999, line 243: dtype is currently the only part of self that overrides whats coming in.... Do we want the self to override what comes in from other in the rest of the attributes?

karlhigley: I think if we're using `__or__` we want `self` to have priority over `other` across the board

karlhigley (changes requested):
Since we're using `|`, we should make `self` have priority over `other` in the merge, so that it matches the semantics of other `or`/`union` operations that `|` represents

merlin/schema/schema.py
jperez999: We do not want this. This actually breaks how we create ancestor schema in the dag. Unless we want to change how that works inside the merge logic in schema. This change alone causes breaks in the nvtabular tests. Particularly in the test_dask_nvt tests.

jperez999: the following line of code starts delivering the original dtype for the column instead of the latest dtype after transformations occur. https://github.com/NVIDIA-Merlin/core/blob/644d4e302f6cd27f6392cc9d862be7b18425cb1d/merlin/dag/node.py#L242

karlhigley: Yeah, in order for `self` having priority to work, we'd have to flip the order of the arguments where it's used too. I think there's only one place we use `__merge__` though so hopefully that's not too bad?

karlhigley (approved):

merlin/schema/schema.py
```diff
@@ -297,13 +297,17 @@ def value_count(self) -> Optional[Domain]:
         value_count = self.properties.get("value_count")
         return Domain(**value_count) if value_count else None
 
-    def __merge__(self, other):
+    def __or__(self, other):
```
oliverholworthy, line 300: Is this roughly equivalent in purpose to the `__add__` on the Schema level? if so, maybe would be clearer to align both the Schema and ColumnSchema with a similar API for combining?

tests/unit/schema/test_column_schemas.py
```diff
@@ -296,3 +296,32 @@ def test_setting_flags_updates_shape_and_value_counts():
...
+    )
+    col_schema_b = ColumnSchema(
+        "col_b",
+        dtype=md.float64,
+        is_list=True,
+        is_ragged=False,
+        properties={"value_count": {"min": 5, "max": 5}},
+    )
+
+    new_schema = col_schema_b | col_schema_a
```
oliverholworthy, line 313: I wonder if matching the behaviour of Python dict union would provide a more familiar experience for users of this operator?

> If a key appears in both operands, the last-seen value (i.e. that from the right-hand operand) wins
> https://peps.python.org/pep-0584/

karlhigley: Oh interesting, that's a fair point. I suggested the other order because it matches the semantics of `or` where the first argument is returned unless it's false-y, otherwise the second.

merlin/schema/schema.py
oliverholworthy: >  addition doesn't imply that either side gets priority in the same way that or/union do.

The implementation of Schema addition calls this ColumnSchema merge method, which suggests that the current sche,a addition does depend on the order. Making it a union operator would be more consistent if that's what it's intended to do.

karlhigley: Agreed

karlhigley:
@jperez999 @oliverholworthy I'm good with either decision we could make about precedence, but matching the PEP-584 order seems like it could be good. I'll defer to the two of you to reach an agreement on which way to do it.

PR #204, opened 2023-02-02
Fix for updated versions of cudf to parquet

jperez999:
This PR makes it possible to use newer versions of cudf (22.12+). In the new versions to_parquet no longer allows the arbitrary passing of kwargs. So we changed chunk_size to row_group_size_rows. At the minimum rows that is allowed is 5000


karlhigley (approved):

karlhigley:
It doesn't look like this works without also bumping versions?

PR #205, opened 2023-02-02
Clean up of make_df function

jperez999:
This PR combines increases readability of the make_df function while accounting for additional possibilities of input explicitly (cupy array being turned into a pandas df). We need this so that we do not hit the following error anymore when turning a GPU ndarray into a pandas DF: `TypeError: Implicit conversion to a NumPy array is not allowed. Please use `.get()` to construct a NumPy array explicitly.`


merlin/core/dispatch.py
```diff
@@ -489,16 +496,21 @@ def concat(objs, **kwargs):
...
-        return pd.DataFrame(_like_df)
-    elif isinstance(_like_df, (cudf.DataFrame, cudf.Series)):
+    if not cudf or device == "cpu" or isinstance(_like_df, (pd.DataFrame, pd.Series)):
+        # move to pandas need it on CPU (host memory)
+        # can be a cudf, cupy or numpy Series
+        if cudf and isinstance(_like_df, (cudf.DataFrame, cudf.Series)):
+            # move to cpu
+            return _like_df.to_pandas()
+        if cp and isinstance(_like_df, cp.ndarray):
+            return pd.DataFrame(cp.asarray(_like_df))
```
oliverholworthy, line 506: `cupy.asarray` returns a cupy array which might not work well with the pandas DataFrame contructor. An additional test for what is expected would help clarify what we'd like to convert from/to.

merlin/core/dispatch.py
```diff
@@ -47,9 +45,18 @@
             # cudf < 21.08
             from cudf.utils.dtypes import is_list_dtype as cudf_is_list_dtype
             from cudf.utils.dtypes import is_string_dtype as cudf_is_string_dtype
-
     except ImportError:
-        HAS_GPU = False
+        ...
```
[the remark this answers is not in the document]
jperez999: yes, there can be... Suppose your on a system that has cupy but does not have cudf. In the current method you find yourself not able to import cupy because cudf failed. And since we dont really require cudf for the use of Systems and this package is a base component for that. I dont think we should be grouping those packages all together. If a package is available it should be able to import.

merlin/core/dispatch.py
jperez999: Actually that particular is the only way you can avoid the `TypeError: Implicit conversion to a NumPy array is not allowed. Please use .get() to construct a NumPy array explicitly.` When translating from cupy to pandas. Good call on the test though, I will put in a test to illustrate that.

merlin/core/dispatch.py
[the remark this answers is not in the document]
jperez999: I guess I misunderstood your comment initially. I think you are highlighting that it doesn't need to be part of this PR. I can see that. I will make a separate PR for those changes. But the motivation behind them, I think, is still valid.

jperez999:
The failures in nvtabular come from the shape changes that were introduced.

tests/unit/core/test_dispatch.py
```diff
@@ -44,3 +51,15 @@ def test_concat_columns(device):
     data_frames = [df1, df2]
     res = concat_columns(data_frames)
     assert res.columns.to_list() == ["a", "b", "c"]
+
+
+@pytest.mark.skipif(not cp, reason="Cupy not available")
+def test_pandas_cupy_combo():
```
oliverholworthy, line 57: is there intended to be a call to the `make_df` function in this test?

jperez999: Originally the idea was to show that the code in make_df was correct. But I will add the make_df call here also. And compare both DFs.

oliverholworthy (approved):

PR #206, opened 2023-02-02
Rely on numba only for gpu detection

jperez999:
This PR simplifies the requirements for detecting whether or not we are on a GPU system. This is part of an effort to ensure that GPU libraries are not loaded when a GPU is not detected, even if they are installed in the environment.

### DIFF
--- a/merlin/core/compat.py
+++ b/merlin/core/compat.py
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import numba
+
 try:
     from numba import cuda
 
@@ -20,5 +22,8 @@ except ImportError:
     cuda = None
 
 HAS_GPU = False
-if cuda and len(cuda.gpus) > 0:
-    HAS_GPU = True
+try:
+    if cuda and len(cuda.gpus) > 0:
+        HAS_GPU = True
+except numba.cuda.CudaSupportError:
+    ...
--- a/merlin/core/compat.py
+++ b/merlin/core/compat.py
@@ -20,12 +20,5 @@ except ImportError:
     cuda = None
 
 HAS_GPU = False
-try:
-    from dask.distributed.diagnostics import nvml
-
-    HAS_GPU = nvml.device_get_count() > 0
-except ImportError:
-    # We can use `cuda` to set `HAS_GPU` now that we
-    # know `distributed` is not installed (otherwise
-    # the `nvml` import would have succeeded)
-    HAS_GPU = cuda is not None
+if cuda and len(cuda.gpus) > 0:
+    HAS_GPU = True


karlhigley:
@oliverholworthy I don't remember exactly what the reason was, but I'm pretty sure I remember some previous discussion between you and @rjzamora about why the GPU detection was written this way. Thoughts on this change?

merlin/core/compat.py
```diff
@@ -21,11 +23,7 @@
...
-    HAS_GPU = nvml.device_get_count() > 0
-except ImportError:
-    # We can use `cuda` to set `HAS_GPU` now that we
-    # know `distributed` is not installed (otherwise
-    # the `nvml` import would have succeeded)
-    HAS_GPU = cuda is not None
+    if cuda and len(cuda.gpus) > 0:
+        HAS_GPU = True
+except numba.cuda.CudaSupportError:
+    ...
```
rjzamora, line 29: See: https://github.com/NVIDIA-Merlin/core/issues/144

(Sorry - I thought the link to that issue was included in this comment).  I believe it may still be a problem to call `cuda.gpus` here if you plan to deploy a dask-cuda cluster.

rjzamora: good catch @karlhigley

oliverholworthy: As @rjzamora mentioned above, the use of NVML instead of numba.cuda was previously made because of the side-effects of using the cuda API before Dask-CUDA spawns processes.

I've opened a PR to add a comment for future reference https://github.com/NVIDIA-Merlin/core/pull/207

There is one edge case I think that isn't handled by this code currently. If you were to set the cuda env var `CUDA_VISIBLE_DEVICES=""` to an empty string. The NVML API will report the same number of devices as before, while `numba.cuda.gpus` would raise a `CudaSupportError` exception.  @jperez999 is this the scenario you had in mind. Or is there something else that we're not handling correctly?

The case where a subset of the GPUs is made visible with `CUDA_VISIBLE_DEVICES` is handled ok since we only need to know if _at least one_ GPU is available for setting this variable

rjzamora: In the future, the nvutil will probably end up being the recommended library (across rapids, etc) to check for things like GPU availability and visible devices - cc @pentschev

jperez999: Yeah so this work is to make it possible to run our GPU containers without using the gpu flag. And ensure everything still runs smoothly. You can also find yourself in a state where you do have pynvml and then using that context fails automatically. Maybe the answer is to make sure pynvml is in the requirements (https://github.com/NVIDIA-Merlin/core/blob/main/requirements.txt) I found myself in a scenario where I installed everything from pip and even though I had GPUs it was reporting that I didnt, because it couldnt run that command successfully. But cupy was able to load and that causes issues like `TypeError: Implicit conversion to a NumPy array is not allowed. Please use `.get()` to construct a NumPy array explicitly.`  when converting from cupy to pandas. A secondary effect I was trying to achieve was to make it frictionless to move between CPU and GPU when your trying to serve a workflow on triton (which may be a bigger lift) considering how it is saved.

pentschev: NVML is _NOT_ affected by `CUDA_VISIBLE_DEVICES`. In Dask-CUDA we have code to deal with those cases, e.g., [matching NVML and `CUDA_VISIBLE_DEVICES`](https://github.com/rapidsai/dask-cuda/blob/84f4aa2e73b28aef0139cb88a83459e3def08b4b/dask_cuda/utils.py#L546-L586). However, I agree that dealing with those differences quickly becomes cumbersome, and that's why, as Rick mentioned above, I started working on a library (currently called NVUtil) to simplify those cases. It's only in its infancy yet and is not public yet, but I hope to move all this sort of specialized code out of Dask to a place that's more accessible by other projects as well.

I'm still not sure I follow what exactly are the cases this PR is trying to solve, but if it's only to figure whether GPUs are available on the system, you could still use NVML and check how many GPUs are available. If you want to respect `CUDA_VISIBLE_DEVICES`, then you may want to copy the function I linked above to make things work correctly. Numba is limited in what it can provide, for example querying MIG devices is not available with Numba, plus you always risk accidentally creating a CUDA context before spawning Dask workers which is problematic as well, so I would suggest to continue using NVML.

jperez999:
Will not merge. Incorrect solution.

PR #225, opened 2023-02-21
Update `concat_columns` to ignore empty DataFrames

oliverholworthy:
Relates to the issue here with dtype mis-matches: https://github.com/NVIDIA-Merlin/NVTabular/issues/1767

This dtype coercision is happening in a pandas concat operation.

There is logic in the DAG executor where we use the function `merlin.core.distpatch.concat_columns` and sometimes pass an empty dataframe as the second value. However this empty dataframe has an index which is longer than the first dataframe. This results in a dtype coercion where pandas converts ints to floats.

```python
pd.concat([pd.DataFrame({"a": [1]}), pd.DataFrame(index=[0])])
# =>
     a
0  1.0
0  NaN
```

This type coercsion is different in cudf.

```python
import pandas as pd
import cudf
from merlin.core.dispatch import concat_columns

# -------------------------------------------------
# pandas

input_data = pd.DataFrame({"session": ["a"], "category": [1]})
parent_data = pd.DataFrame({"session": ["a", "b"]})

concat_columns([input_data, parent_data[[]]])
# =>
  session  category
0       a       1.0
1     NaN       NaN

# -------------------------------------------------
# cudf

input_data = cudf.DataFrame({"session": ["a"], "category": [1]})
parent_data = cudf.DataFrame({"session": ["a", "b"]})

concat_columns([input_data, parent_data[[]]])
# =>
  session category
0       a        1
1    <NA>     <NA>
```

### DIFF
--- a/merlin/core/dispatch.py
+++ b/merlin/core/dispatch.py
@@ -358,15 +358,15 @@ def concat_columns(args: list):
     if len(args) == 1:
         return args[0]
     elif cudf is not None and isinstance(args[0], cudf.DataFrame):
-        return cudf.concat(
-            [a.reset_index(drop=True) for a in args if not a.empty],
-            axis=1,
-        )
+        objs = [a.reset_index(drop=True) for a in args if not a.empty]
+        if objs == []:
+            return cudf.DataFrame()
+        return cudf.concat(objs, axis=1)
     elif isinstance(args[0], pd.DataFrame):
-        return pd.concat(
-            [a.reset_index(drop=True) for a in args if not a.empty],
-            axis=1,
-        )
+        objs = [a.reset_index(drop=True) for a in args if not a.empty]
+        if objs == []:
+            return pd.DataFrame()
+        return pd.concat(objs, axis=1)
     elif isinstance(args[0], DictLike):
         result = type(args[0])()
         for arg in args:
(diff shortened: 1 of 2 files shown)


karlhigley (approved):

merlin/core/dispatch.py
```diff
@@ -359,12 +359,12 @@ def concat_columns(args: list):
...
     elif cudf is not None and isinstance(args[0], cudf.DataFrame):
         return cudf.concat(
-            [a.reset_index(drop=True) for a in args],
+            [a.reset_index(drop=True) for a in args if not a.empty],
             axis=1,
         )
     elif isinstance(args[0], pd.DataFrame):
         return pd.concat(
-            [a.reset_index(drop=True) for a in args],
+            [a.reset_index(drop=True) for a in args if not a.empty],
```
karlhigley, line 367: From the NVT test failures, it looks like this `concat` call fails if there are no dataframes to concatenate, so the list comprehension might need to move out to a local that can be checked before calling `concat`

oliverholworthy: I've updated with this logic, and still have an error. It seems ignoring empty dataframes is maybe not the correct thing to do in the general case. NVTabular is relying on concating empty dataframes where the index is preserved (concated).

oliverholworthy:
Closing this because the original issue can be fixed by wrapping the call to this `concat_columns` in the executor.

And for now, it's not clear what logic should be applied here in the case where one or more dataframes is empty

oliverholworthy:
alterntative solution to the NVTabular issue (https://github.com/NVIDIA-Merlin/NVTabular/issues/1767 ) that motivated this PR proposed in #226

PR #230, opened 2023-02-24
Introduce TensorTable, a cross framework abstraction

jperez999:
This PR introduces TensorTable, a cross framework abstraction. That provides the ability to move data between different supported frameworks (i.e. cupy, np, torch, tensorflow) using zero copy mechanism via dlpack.


merlin/table/tensor_column.py
```diff
@@ -0,0 +1,155 @@
...
+from merlin.core.protocols import SeriesLike
+
+
+class Device(Enum):
+    CPU = 0
+    GPU = 1
+
+
+# This should always contains arrays or tensors, not series
+class TensorColumn(ABC, SeriesLike):
```
karlhigley, line 31: `TensorColumn` should be `SeriesLike`, but I don't think we can explicitly declare it here and still appease the linter. Maybe we could add tests that check that `isinstance(TensorColumn(), SeriesLike)` and `isinstance(TensorTable(), DataFrameLike)`?

merlin/table/tensor_column.py
```diff
@@ -0,0 +1,155 @@
...
+            array, expected_type
+        ):
+            raise TypeError(
+                f"{self.__class__} expected arrays of type {expected_type}, "
+                f"but received {field_name} of type {type(array)}"
+            )
+
+
+@dataclass(frozen=True)
+class TransferColumn:
```
karlhigley, line 127: Wonder if this class and the sub-classes below should be semi-private and have names starting with leading underscores?

merlin/table/tensor_table.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def __setitem__(self, key, value):
+        self._columns[key] = value
+
+    def __delitem__(self, key):
+        del self._columns[key]
+
+    @property
+    def device(self):
+        return list(self.values())[0].device
```
oliverholworthy, line 67: it seems like it's currently possible to construct a TensorTable with columns of different types and potentially on different devices. If that's something we'd like to support then this device property may need to change. to return for example a dictionary of devices by column name?

karlhigley: I think we should add validation to prevent that 👍🏻

merlin/core/dispatch.py
```diff
@@ -715,3 +715,46 @@ def get_random_state():
...
+        if is_list_dtype(df[col]):
+            values_series, offsets_series = pull_apart_list(df[col], device=device)
+            array_cols[col] = column_type(values_series.values, offsets_series.values)
+        else:
+            array_cols[col] = column_type(df[col].values)
+
+    return TensorTable(array_cols)
+
+
+def df_from_tensor_table(table):
```
oliverholworthy, line 737: If we want to support TensorTable conversions to DataFrames, Would it be easier to discover this functionality for users by making a call to this from a method like `tensor_table.to_dataframe()`?

karlhigley: Yeah, it would be easier to discover that way. One of the wrinkles that led us down this path is that we didn't want to make `TensorTable` itself depend on Pandas/cuDF`. `dispatch.py` is where we currently handle operations that depend on dataframe types and whether or not `cudf` is available, so we went with that for now. I suppose we could put the dataframe library imports directly in these methods to avoid importing them everywhere `TensorTable` is used.

oliverholworthy: adding the imports into the methods would be a reasonable option if we wanted to add these methods. We could do that in a follow-up PR. If we added a `to_dataframe` method, it would also be nice to have a corresponding `from_dataframe` classmethod

merlin/table/tensor_table.py
oliverholworthy: Adding a restriction on the same device seems reasonable. What about tensors of different types that are on the same device, should we restrict not only to the same device but also to the same column type?

karlhigley: Yeah, I think so. I've added both validations in the latest commits

merlin/core/dispatch.py
karlhigley: That's the style I like too. 👍🏻 

An open question for me is how we want the methods for converting across frameworks at the table level to look. With dtype conversion we did `dtype.to("numpy")`, which works well when some types or frameworks may not be available but has the downside of being "stringly typed." Unless we come up with something better, I'd probably use that here too, but don't feel entirely great about it.

merlin/table/tensor_column.py
jperez999: Ok I went ahead and made this change. I also removed all of the array based logic, since we are using dlpack only.

tests/unit/table/test_tensor_table.py
```diff
@@ -0,0 +1,290 @@
...
+    source_column_type, source_col_constructor, array_lib = source_column
+    target_column_type, target_input = target_column
+    op = PaddingOperator(3, array_lib=array_lib)
+    expected_output = source_col_constructor([[1, 0, 0], [2, 3, 0]])
+
+    tensor_table = TensorTable(target_input)
+
+    # Column conversions would happen in the executor
+    for col_name, column in tensor_table.items():
+        tensor_table[col_name] = convert_col(column, source_column_type)
```
oliverholworthy, line 203: is this checking that converting a column that's already of the target type a no-op? (is this required for the test to pass?

tests/unit/table/test_tensor_column.py
```diff
@@ -0,0 +1,71 @@
...
+    assert all(np_col.values == values)
+
+
+def test_dtype():
+    values = np.array([1, 2, 3, 4, 5, 6, 7, 8])
+    np_col = NumpyColumn(values=values)
+    assert np_col.dtype == md.dtype(values.dtype)
+
+    np_col = NumpyColumn(values=values, offsets=np.array([0, 2, 4, 6, 8]))
+    assert np_col.dtype == md.dtype(values.dtype)
```
oliverholworthy, line 58: this might be something for a future PR. The shape of the underlying data not currently captured in the dtype, and could be useful to be able to access from this merlin dtype, particularly for ragged columns where we've specified the offsets

merlin/core/dispatch.py
oliverholworthy: If the framework column classes are intended to be part of the public API, then could use those instead of strings? something along the lines of `table.convert_columns_to(TorchColumn)` or maybe something more concise. Since we don't need this functionality in the immediate term in systems we can figure this out in a follow-up change

merlin/table/tensor_column.py
```diff
@@ -0,0 +1,157 @@
...
+
+        self._values = values
+        self._offsets = offsets
+
+        self._dtype = md.dtype(dtype or values.dtype)
+        self._ref = _ref
+        self._device = _device
+
+    @property
+    def device(self) -> Device:
```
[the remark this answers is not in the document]
oliverholworthy: maybe a separate `device_type` and `device_index` would be an option?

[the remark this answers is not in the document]
karlhigley: Yeah, I like the idea of a device index method. We don't as of yet have a reason to worry about it too much, but that's the route I'd see us taking.

tests/unit/table/test_tensor_table.py
karlhigley: It's not required for the test to pass

tests/unit/table/test_tensor_column.py
karlhigley: Yeah, capturing the shapes would be good, agreed. `DictArray` didn't have that functionality so I don't think it's crucial yet, but we definitely do want to have it here in the longer term.

karlhigley:
Further work surfaced here:
* Adding cross-framework conversion methods at the table level
* Adding shapes to TensorColumn dtypes
* Adding a `device_index` method to differentiate between CPUs/GPUs

I think we can move forward with this PR and add those incrementally as we go though.

oliverholworthy (approved):
It feels like this is in a good place to start trying this out in Systems and figure out if further refinement is needed from that experience.

There are some open questions about how the methods for converting to/from dataframes and frameworks should work. I think we're ok to figure that out later as the use-cases arise.

PR #231, opened 2023-03-01
Add shapes to `TensorColumn`s

karlhigley:
This constructs a Merlin shape from the values and offsets provided to the `TensorColumn` constructor, which differs from the underlying framework array/tensor shapes since a single column may be represented by multiple arrays/tensors. The tests check that shape construction works across Numpy, Cupy, Tensorflow, and Torch.

### DIFF
--- a/merlin/table/tensor_column.py
+++ b/merlin/table/tensor_column.py
@@ -102,11 +102,18 @@ class TensorColumn:
     def dtype(self):
         return self._dtype
 
+    def __len__(self):
+        if self.offsets is not None:
+            return len(self.offsets) - 1
+        else:
+            return len(self.values)
+
     def __getitem__(self, index):
         # get correct indexes if offsets exists
         if self._offsets is None:
             index = len(self._values) + index if index < 0 else index
             return self._values[index]
+
         # There should be a better way to get negative indexing to work
         index = len(self._offsets) - 1 + index if index < 0 else index
         start = self._offsets[index]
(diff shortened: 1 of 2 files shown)


merlin/table/tensor_column.py
```diff
@@ -108,6 +130,16 @@ def __eq__(self, other):
             and self.device == other.device
         )
 
+    def _construct_shape(self, values, offsets):
+        if offsets is not None:
+            num_rows = len(offsets) - 1
+            row_lengths = offsets[1:] - offsets[:-1]
+            num_cols = int(row_lengths[0]) if all(row_lengths == row_lengths[0]) else None
+            shape = Shape((num_rows, num_cols))
```
oliverholworthy, line 138: would we like to support 2-dimensional `values` here (e.g. corresponding to arrays in a 3-d ragged column). For example a ragged list column containing lists of fixed-size embeddings.

for example a column like (shape: (2, None, 2) ):

```
    [[1, 2], [3, 4]]
            [[3, 5]]
```

with `values` (shape: (3, 2) ):

```
[
  [1, 2],
  [3, 4],
  [3, 5]
]
```

and `offsets` (shape: (3,) ):

```
[0, 2, 3]
```

Could be an enhancement for later if we don't have a reqirement for this now

oliverholworthy (approved):

merlin/table/tensor_column.py
karlhigley: We would like to support those, which @jperez999 is starting to take a look at. I like this thought about how to encode the values in a fixed shape tensor—let's try it out and see if it works for our use cases. 👍🏻 

I think I'd opt to leave it for a future PR though, since we only need support for 2-d shapes in Systems (which is the current `TensorTable` test bed) for now.

PR #233, opened 2023-03-07
Add `TensorTable.to/from_df()` methods

karlhigley:
Although the functionality already existed in the form of dispatch functions, this provides some additional convenience when converting between tensor tables and dataframes. It also adjusts the dictionary conversions to match this style.

### DIFF
--- a/merlin/table/tensor_table.py
+++ b/merlin/table/tensor_table.py
@@ -15,8 +15,8 @@
 #
 from typing import Any, Dict
 
-from merlin.core.dispatch import df_from_tensor_table, tensor_table_from_df
 from merlin.dag.utils import group_values_offsets
+from merlin.table.conversions import df_from_tensor_table, tensor_table_from_df
 from merlin.table.cupy_column import CupyColumn
 from merlin.table.numpy_column import NumpyColumn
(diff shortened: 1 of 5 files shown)


merlin/table/tensor_table.py
```diff
@@ -31,6 +32,10 @@ class TensorTable:
     A DataFrameLike wrapper around a dictionary of arrays or tensors
     """
 
+    @classmethod
+    def from_df(cls, df):
+        return tensor_table_from_df(df)
```
oliverholworthy, line 37: In the discussion on the PR that introduced TensorTable there was a mention of inline imports to avoid an import of dataframe libs when using this, is that no longer required here?

karlhigley: I still think it's a good idea to decouple this from dataframe libraries as much as possible, but my thinking on how to do that has shifted a bit. Since Pandas and cuDF require NumPy and CuPy (respectively), I think we can convert these functions to use `lazy_singledispatch` (because the availability of the dataframe types also implies availability of the relevant array library.)

I'll take a stab at that and add it to this PR.

oliverholworthy (approved):

PR #236, opened 2023-03-08
Support `Dataset` cpu-mode in environment with GPUs that have not been detected 

oliverholworthy:
Support `Dataset` cpu-mode in environment with GPUs that have not been detected.

When CUDA visible devices environment variable is empty `CUDA_VISIBLE_DEVICES=""`

Currently you'll get an error like the following in the case where `cudf` has imported sucessfully while `HAS_GPU` is False.

```
AttributeError: 'NoneType' object has no attribute 'DataFrame'
```


merlin/core/dispatch.py
```diff
@@ -217,50 +213,78 @@ def pd_convert_hex(x):
...
+        )
 
 
 def arange(size, like_df=None, dtype=None):
     """Dispatch for numpy.arange"""
-    if not HAS_GPU or isinstance(like_df, (np.ndarray, pd.DataFrame, pd.Series)):
+    if isinstance(like_df, (np.ndarray, pd.DataFrame, pd.Series)):
         return np.arange(size, dtype=dtype)
-    else:
+    elif cudf and isinstance(like_df, (cp.ndarray, cudf.DataFrame, cudf.Series)):
```
karlhigley, line 232: The code in dispatch.py was written in this (admittedly somewhat weird) fashion to avoid directly referencing GPU-only libraries that won’t be available on CPU. I think the path forward here is likely to apply ‘@lazy_single dispatch’ so that we register a GPU version of the function only when CuPy/cuDF are available

karlhigley (commented):
Everything outside dispatch.py LGTM, but I have concerns about using the GPU classes there

merlin/core/dispatch.py
oliverholworthy: It turned out to be easier to keep this PR representing the addition of a test for GPU environment with no visible devices (and corresponding fixes to the tensor table tests). Moved the dispatch changes to #244 .

tests/unit/table/test_tensor_column.py
```diff
@@ -91,7 +92,7 @@ def test_equality():
     assert np_col != np_col_3
 
 
-@pytest.mark.skipif(cp is None, reason="requires GPU")
+@pytest.mark.skipif(not (cp and HAS_GPU), reason="requires GPU")
```
karlhigley, line 95:
```suggestion
@pytest.mark.skipif(not (cp and HAS_GPU), reason="requires CuPy and GPU")
```

tests/unit/table/test_tensor_column.py
```diff
@@ -108,7 +109,7 @@ def test_cupy_cpu_transfer():
     assert isinstance(cpu_col_again, NumpyColumn)
 
 
-@pytest.mark.skipif(cp is None, reason="requires GPU")
+@pytest.mark.skipif(not (cp and HAS_GPU), reason="requires GPU")
```
karlhigley, line 112:
```suggestion
@pytest.mark.skipif(not (cp and HAS_GPU), reason="requires CuPy and GPU")
```

merlin/core/dispatch.py
oliverholworthy: I also added another piece of information in the description of that PR, which is that prior to #211 , the `HAS_GPU` variable represented both the detection of a GPU device and the successful import of cudf. Assuming a successful import of cudf tells us both whether cudf is installed and whether or not there is at least one GPU device then the checking of the cudf  before every usage is roughly equivalent to what we were doing prior to #211

karlhigley: Ah yeah, I hadn't considered that checking `cudf` also implicitly checks `cupy` 👍🏻

oliverholworthy: > the isinstance check uses both cudf and cp but only checks if cudf exists first

If cudf depends on cupy, then this might be ok, since the presence of cudf implies cupy is also available?

karlhigley: Agreed ✅

tests/unit/table/test_convert_column.py
```diff
@@ -27,7 +28,7 @@
 source_cols: List[TensorColumn] = []
 output_col_types: List[Type] = []
 
-if cp:
+if cp and HAS_GPU:
```
karlhigley, line 31: Is there a way we could put the `HAS_GPU` in `merlin.core.compat` so we don't have to repeat it in so many places?

karlhigley: I think so? Can we import CuPy successfully if we don't have a GPU?

oliverholworthy: that depends on the extent of what "don't have a GPU" means

If CuPy or cudf fail to find relevant cuda drivers they raise an ImportError which we catch in compat. and `cp` will be None. However, if we set `CUDA_VISIBLE_DEVICES=` in an environment that has the relevant cuda drivers and libs available then the import is successful, `cp` will have the value of the cupy module, but if you try to use it e.g. cupy.array(...), then you get a `CUDARuntimeError`

oliverholworthy: Is it possible this is a misuse of CUDA_VISIBLE_DEVICES and not a scenario we would like to support.

karlhigley: In the latter case, is `HAS_GPU` false? If so, then seems like we might be able to use it in compat

karlhigley:
This test environment looks like it breaks some of the Dask cluster tests

karlhigley (approved):

PR #243, opened 2023-03-13
Check environment supports target device in Dataset constructor

oliverholworthy:
Add checks of the environment to make sure the target device is supported based on the `cpu` parameter.

## Current behaviour

`cpu=False` is treated as equivalent to `cpu=None`.

- `cpu=None` or `cpu=False`
  - will use GPU device if cudf available, otherwise host CPU
 - cpu = True
   - use CPU host device 

## After this PR

`cpu=False` is changed to mean we'd like to use the GPU device only and if for some reason we can't, we'd like to know about this and why. 

- cpu = None
  - will use GPU device if available
- cpu = False
  - use GPU device
   - raises RuntimeError if either device not detected with pynvml or cudf is not available.
 - cpu = True
   - use CPU host device 

## Motivation

Making device specified for Dataset clearer. 

Introducing earlier failure in the case of missing cudf or device availability reduces the chance of subsequent later on when we try to use functions in `merlin.core.dispatch` that then fail with an error that may be less clear.


karlhigley:
This LGTM, but wonder if it’s worth adding some tests? They wouldn’t all be able to run in every environment but between the CPU and GPU CI, we should be able to test the four (?) possibilities.

karlhigley:
I can't completely tell if the downstream failures are related, but they look like they could be

oliverholworthy:
> I can't completely tell if the downstream failures are related, but they look like they could be

Yes, the errors in NVTabular/dataloader are related. There are two more places that need to change (both are from test code where we pass cpu=False in environment without GPUs):

- https://github.com/NVIDIA-Merlin/dataloader/pull/114
- https://github.com/NVIDIA-Merlin/NVTabular/pull/1787

karlhigley:
Looks like there are still some downstream issues in NVT and the dataloaders, but less than there used to be 👍🏻

karlhigley (approved):

PR #244, opened 2023-03-13
Remove use of HAS_GPU from `dispatch` functions

oliverholworthy:
Remove use of HAS_GPU from `dispatch` functions. 

Following #211 the variable `HAS_GPU` no longer includes the information about whether or not cudf is installed. This can result in confusing errors when using the dispatch functions in a GPU enabled environment without cudf available. (`'NoneType' object has no attribute 'DataFrame'`)

### DIFF
--- a/merlin/core/compat.py
+++ b/merlin/core/compat.py
@@ -62,6 +62,11 @@ except ImportError:
     cupy = None
 
 try:
+    import cudf
+except ImportError:
+    cudf = None
+
+try:
     import tensorflow
     from tensorflow.python.framework import ops as tf_ops
 except ImportError:
--- a/merlin/core/dispatch.py
+++ b/merlin/core/dispatch.py
@@ -24,16 +24,16 @@ import pandas as pd
 import pyarrow as pa
 import pyarrow.parquet as pq
 
-from merlin.core.compat import HAS_GPU
+# unused HAS_GPU import is here for backwards compatibility
+from merlin.core.compat import HAS_GPU  # pylint: disable=unused-import # noqa: F401
+from merlin.core.compat import cupy as cp
 from merlin.core.protocols import DataFrameLike, DictLike, SeriesLike
 
 cudf = None
-cp = None
 rmm = None
 
-if HAS_GPU:
+if cudf:
     try:
-        import cudf  # type: ignore[no-redef]
         import dask_cudf
         import rmm  # type: ignore[no-redef]
         from cudf.core.column import as_column, build_column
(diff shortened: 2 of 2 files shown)


merlin/core/dispatch.py
```diff
@@ -24,16 +24,16 @@
...
+from merlin.core.compat import HAS_GPU  # pylint: disable=unused-import # noqa: F401
+from merlin.core.compat import cupy as cp
 from merlin.core.protocols import DataFrameLike, DictLike, SeriesLike
 
 cudf = None
-cp = None
 rmm = None
 
-if HAS_GPU:
+if cudf:
```
karlhigley, line 35: I could be missing something, but seems like `cudf` would need to imported from `merlin.core.compat` for this to work

oliverholworthy: I missed this commit from moving from the other PR. Updated

karlhigley (approved):

merlin/core/compat.py
```diff
@@ -61,6 +61,11 @@ def _get_gpu_count():
 except ImportError:
     cupy = None
 
+try:
+    import cudf
```
oliverholworthy, line 65: Based on your comment here ("[may blow up in CPU-only environments](url)") @karlhigley  

This is currently the case if you try to import cudf in our containers without GPUs available, you'll get a Segmentation fault, which the except ImportError can't catch

```
Exception ignored in: 'cuda._lib.ccudart.utils.cudaPythonGlobal.lazyInitGlobal'
Traceback (most recent call last):
  File "cuda/_cuda/ccuda.pyx", line 3671, in cuda._cuda.ccuda._cuInit
  File "cuda/_cuda/ccuda.pyx", line 435, in cuda._cuda.ccuda.cuPythonInit
RuntimeError: Failed to dlopen libcuda.so
Segmentation fault (core dumped)
```

oliverholworthy: The segmentation fault was fixed in cuda-python 11.7.1 , however that requires use of cudf 22.12 and above. With that in mind I think it might be ok to keep the import as it is and say that support for CPU-only environments that have cudf installed requries cudf >=22.12 ?

karlhigley: Yeah, that sounds good. You're way ahead of me (as usual.)

karlhigley:
It looks like some of the NVT test failures might actually be related to these changes

oliverholworthy:
> It looks like some of the NVT test failures might actually be related to these changes

I've updated and resolved these issues. The typehint for `read_dispatch` suggested that it might only accept a dataframe type, but is also called with strings. Updated to handle that case. 

And the functions `{random_state, arange, array, zeros}` all had the same pattern. Updated to handle RangeIndex type passed as `like_df` and the base case where like_df is None and we're simply picking between cupy and numpy. The code is more verbose, but is clearer about what it's doing now. Hopefully a further refactor later will remove the need for these altogether

PR #249, opened 2023-03-16
Change the location and filename of schema.pbtxt to .merlin/schema.json

edknv:
Fixes https://github.com/NVIDIA-Merlin/Merlin/issues/845.

This PR moves Merlin-specific metadata files, such as schema, to a sub-directory `.merlin` in an effort to standardize the location and file format of Merlin schema. The format of the schema file is also changed to JSON.

`Models Tests (CPU)` in the GitHub Actions CI is failing, but this is expected as Models hosts the dataset tests. PR for addressing this: https://github.com/NVIDIA-Merlin/models/pull/1021

### DIFF
--- a/tests/unit/io/test_io.py
+++ b/tests/unit/io/test_io.py
@@ -91,7 +91,8 @@ def test_dataset_infer_schema(dataset, engine):
 
 @pytest.mark.parametrize("engine", ["csv", "parquet", "csv-no-header"])
 @pytest.mark.parametrize("cpu", [None, True])
-def test_string_datatypes(tmpdir, engine, cpu):
+@pytest.mark.parametrize("file_format", ["pbtxt", "json"])
+def test_string_datatypes(tmpdir, engine, cpu, file_format):
     df_lib = dispatch.get_lib()
     df = df_lib.DataFrame({"column": [[0.1, 0.2]]})
     dataset = merlin.io.Dataset(df)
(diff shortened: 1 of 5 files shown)


tests/unit/io/test_io.py
```diff
@@ -494,11 +500,13 @@ def test_validate_and_regenerate_dataset(tmpdir):
...
-    assert os.path.join(path, "_metadata") in file_list
-    assert os.path.join(path, "_file_list.txt") in file_list
-    assert os.path.join(path, "_metadata.json") in file_list
-    assert len(file_list) == N + 3  # N data files + 3 metadata files
+    data_file_list = glob.glob(os.path.join(path, "*"))
+    assert len(data_file_list) == N + 1  # N data files + '_metadata'
+    assert os.path.join(path, "_metadata") in data_file_list
+    metadata_file_list = glob.glob(os.path.join(path, ".merlin", "*"))
+    assert os.path.join(path, ".merlin", "_file_list.txt") in metadata_file_list
+    assert os.path.join(path, ".merlin", "_metadata.json") in metadata_file_list
```
karlhigley, line 508: @rjzamora This PR looks good to me, but I wanted to run these files by you before we move them. Can these move to the `.merlin` directory without breaking things that depend on them?

edknv: @karlhigley Thanks for catching this. I think these are metadata files for hugeCTR, and I was maybe thinking hugeCTR is also Merlin so... After double-checking some of hugeCTR code, I think moving those files might break hugeCTR. I reverted those changes in [a4c7cef](https://github.com/NVIDIA-Merlin/core/pull/249/commits/a4c7ceffe1a1d7473891b58e11236e4e3c804f06) to be safe so that these are still in the same directory as the parquet files, and only `schema.json` goes into `.merlin`.

karlhigley: I agree with you that in the longer term they probably should get consolidated into the Merlin metadata directory, but this does seem safer for now 🤝

karlhigley (approved):

PR #256, opened 2023-03-28
Remove warning about compound tags deprecation

oliverholworthy:
Remove warning about compound tags deprecation. Since the warning message here is not currently easily actionable by users. And prints a message every time a ColumnSchema is created, resulting in distracting messages in logs.

After discussions about this we are considering keeping the compound tags for backwards compatibility for now.

If we reach a point where we're using the narrower tags only across Merlin, we can revisit the deprecation approach for these


oliverholworthy:
Providing a way to simplify selecting by _all_ tags was one of the outcomes. However we also discussed two other things. Keeping the compound tags around as a convenient shortcut and removing the warning. 

The current state of things seems to be that the warning is not actionable. Which suggests that it would likely be clearer to users of our packages if it wasn't printed/logged.

karlhigley (approved):

PR #261, opened 2023-03-29
Run with import without gpu

jperez999:
This PR enables a user to run our GPU docker container without GPUs and successfully be able to run all tests. Previously we had try catches in a few places in the code for importing GPU specific packages. This PR remediates those try catches because in an environment where you have the package but no GPU you end up getting ccuda.pyx init failure. This is because, the package (i.e. cudf, dask_cudf, rmm) do exist but when they try to access information about GPUs it fails and throws an error something like:
```
collecting ... Traceback (most recent call last):
  File "cuda/_cuda/ccuda.pyx", line 3671, in cuda._cuda.ccuda._cuInit
  File "cuda/_cuda/ccuda.pyx", line 435, in cuda._cuda.ccuda.cuPythonInit
RuntimeError: Failed to dlopen libcuda.so
Exception ignored in: 'cuda._lib.ccudart.utils.cudaPythonGlobal.lazyInitGlobal'
Traceback (most recent call last):
  File "cuda/_cuda/ccuda.pyx", line 3671, in cuda._cuda.ccuda._cuInit
  File "cuda/_cuda/ccuda.pyx", line 435, in cuda._cuda.ccuda.cuPythonInit
RuntimeError: Failed to dlopen libcuda.so
Fatal Python error: Segmentation fault
```
This PR leverages the compat file, and makes it the single point of import for the main packages (cudf, cupy) and it adds a security around it that ensures you can only import those packages if GPUs are detected. So if you find yourself in a scenario where the package is installed but no GPUs are detected you can, now, still safely use the core package. Therefore, we can now run our containers with and without the `--gpus=all` flag in docker. This was a customer ask and it helps developers when trying to test cpu only environment on a resource that has GPUs.

### DIFF
--- a/merlin/dtypes/mappings/cudf.py
+++ b/merlin/dtypes/mappings/cudf.py
@@ -15,6 +15,7 @@
 #
 import numpy as np
 
+from merlin.core.compat import cudf
 from merlin.core.dispatch import is_string_dtype
 from merlin.dtypes.mapping import DTypeMapping, NumpyPreprocessor
 from merlin.dtypes.registry import _dtype_registry
@@ -41,17 +42,16 @@ def cudf_translator(raw_dtype) -> np.dtype:
         return category_type
 
 
-try:
-    # We only want to register this mapping if cudf is available, even though
-    # the mapping itself doesn't use cudf (yet?)
+if cudf:
+    try:
+        # We only want to register this mapping if cudf is available, even though
+        # the mapping itself doesn't use cudf (yet?)
 
-    import cudf  # pylint:disable=unused-import # noqa: F401
+        cudf_dtypes = DTypeMapping(
+            translator=NumpyPreprocessor("cudf", cudf_translator, attrs=["_categories"]),
+        )
+        _dtype_registry.register("cudf", cudf_dtypes)
+    except ImportError as exc:
+        from warnings import warn
 
-    cudf_dtypes = DTypeMapping(
-        translator=NumpyPreprocessor("cudf", cudf_translator, attrs=["_categories"]),
-    )
-    _dtype_registry.register("cudf", cudf_dtypes)
-except ImportError as exc:
-    from warnings import warn
-
-    warn(f"cuDF dtype mappings did not load successfully due to an error: {exc.msg}")
+        warn(f"cuDF dtype mappings did not load successfully due to an error: {exc.msg}")
--- a/merlin/io/csv.py
+++ b/merlin/io/csv.py
@@ -16,20 +16,13 @@
 import functools
 
 import dask.dataframe as dd
-
-from merlin.core.compat import cudf
-
-if cudf:
-    try:
-        import dask_cudf
-    except ImportError:
-        dask_cudf = None
-import numpy as np
 from dask.bytes import read_bytes
 from dask.utils import parse_bytes
 from fsspec.core import get_fs_token_paths
 from fsspec.utils import infer_compression
 
+from merlin.core.compat import dask_cudf
+from merlin.core.compat import numpy as np
 from merlin.io.dataset_engine import DatasetEngine
 
 
--- a/merlin/io/csv.py
+++ b/merlin/io/csv.py
@@ -17,10 +17,13 @@ import functools
 
 import dask.dataframe as dd
 
-try:
-    import dask_cudf
-except ImportError:
-    dask_cudf = None
+from merlin.core.compat import cudf
+
+if cudf:
+    try:
+        import dask_cudf
+    except ImportError:
+        dask_cudf = None
 import numpy as np
 from dask.bytes import read_bytes
 from dask.utils import parse_bytes
--- a/merlin/io/dataframe_engine.py
+++ b/merlin/io/dataframe_engine.py
@@ -13,16 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from merlin.core.compat import cudf
-
-if cudf:
-    try:
-        import dask_cudf
-    except ImportError:
-        dask_cudf = None
 from dask.dataframe.core import new_dd_object
 from dask.highlevelgraph import HighLevelGraph
 
+from merlin.core.compat import cudf, dask_cudf
 from merlin.io.dataset_engine import DatasetEngine
 
 
--- a/merlin/io/dataframe_engine.py
+++ b/merlin/io/dataframe_engine.py
@@ -13,11 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-try:
-    import cudf
-    import dask_cudf
-except ImportError:
-    cudf = None
+from merlin.core.compat import cudf
+
+if cudf:
+    try:
+        import dask_cudf
+    except ImportError:
+        dask_cudf = None
 from dask.dataframe.core import new_dd_object
 from dask.highlevelgraph import HighLevelGraph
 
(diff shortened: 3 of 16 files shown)


merlin/table/conversions.py
```diff
@@ -131,8 +185,7 @@ def _tensor_table_from_pandas_df(df: pd.DataFrame):
 
 @tensor_table_from_df.register_lazy("cudf")
 def _register_tensor_table_from_cudf_df():
-    import cudf
-
+    from merlin.core.compat import cudf
```
jperez999, line 188: This import forced me, via interrogate, to add in comments for all public methods in the file.

merlin/io/csv.py
```diff
@@ -17,10 +17,13 @@
...
 
-try:
-    import dask_cudf
-except ImportError:
-    dask_cudf = None
+from merlin.core.compat import cudf
+
+if cudf:
+    try:
+        import dask_cudf
```
karlhigley, line 24: Could this use the `dask_cudf` from `merlin.core.compat`?

merlin/io/dataframe_engine.py
```diff
@@ -13,11 +13,13 @@
...
-try:
-    import cudf
-    import dask_cudf
-except ImportError:
-    cudf = None
+from merlin.core.compat import cudf
+
+if cudf:
+    try:
+        import dask_cudf
```
karlhigley, line 20: Same question here re: using `merlin.core.compat.dask_cudf`

karlhigley (commented):
This seems like a good direction to go in, and I wonder if we can even take it a little farther

oliverholworthy:
> Fatal Python error: Segmentation fault

The segmentation fault was fixed in cuda-python 11.7.1 , however that requires use of cudf 22.12 and above. With that in mind I think it might be ok to keep the import as it is and say that support for CPU-only environments that have cudf installed requries cudf >=22.12 ?

oliverholworthy:
Having the `cudf` imports wrapped with `HAS_GPU` means that we are encapsulating in this the idea of having at least one visible device with the ability to import cudf for example

oliverholworthy:
I've been recently working toward a similar goal in https://github.com/NVIDIA-Merlin/core/pull/244 , https://github.com/NVIDIA-Merlin/core/pull/243 and https://github.com/NVIDIA-Merlin/core/pull/236

oliverholworthy:
Part of those changes involved adidng checks for `HAS_GPU` where relevant alongside checks of `cudf`. At one point `HAS_GPU` was the variable we were using to included the information of both whether we have a GPU, and cudf importable. We split that out in https://github.com/NVIDIA-Merlin/core/pull/211 . And  I suppose this is moving what was once `HAS_GPU` into `compat.cudf`. Which is clearer than what we had before in the `HAS_GPU` variable.

One reason we might consider keeping the compat.cudf separate from compat.HAS_GPU is mostly about edge cases and reporting of errors. [here for example in the Dataset](https://github.com/NVIDIA-Merlin/core/blob/ec9a360485bc48e67bb6fe4b5c5ca092eca2a97c/merlin/io/dataset.py#L249-L261) . And I've seen cases where pynvml fails to initialize for some reason, but cudf is still available. Which may result in some confusion if `cudf` includes this notion.  On the other hand, being able to refer to `compat.cudf` without checking `compat.HAS_GPU` does simplify our checks in tests quite a bit so on balance could be reasonable, if less explicit. And any code that does need to check for cudf availability (not including of visible CUDA devices), we can achieve that with a separate import elsewhere I suppose.

jperez999:
> Having the `cudf` imports wrapped with `HAS_GPU` means that we are encapsulating in this the idea of having at least one visible device with the ability to import cudf for example

I mean, we cant ever use cudf without having a GPU, so that seems like the correct usage. Do you see a time when we can use cudf without a GPU?

jperez999:
> Part of those changes involved adidng checks for `HAS_GPU` where relevant alongside checks of `cudf`. At one point `HAS_GPU` was the variable we were using to included the information of both whether we have a GPU, and cudf importable. We split that out in #211 . And I suppose this is moving what was once `HAS_GPU` into `compat.cudf`. Which is clearer than what we had before in the `HAS_GPU` variable.
> 
> One reason we might consider keeping the compat.cudf separate from compat.HAS_GPU is mostly about edge cases and reporting of errors. [here for example in the Dataset](https://github.com/NVIDIA-Merlin/core/blob/ec9a360485bc48e67bb6fe4b5c5ca092eca2a97c/merlin/io/dataset.py#L249-L261) . And I've seen cases where pynvml fails to initialize for some reason, but cudf is still available. Which may result in some confusion if `cudf` includes this notion. On the other hand, being able to refer to `compat.cudf` without checking `compat.HAS_GPU` does simplify our checks in tests quite a bit so on balance could be reasonable, if less explicit. And any code that does need to check for cudf availability (not including of visible CUDA devices), we can achieve that with a separate import elsewhere I suppose.

So if there is ever a time we cannot detect a GPU, we should not import cudf. Chances are it will also fail inside on internal imports in cudf. We should work to make sure that we detect GPUs and dask nvml should not have problems detecting GPUs. I have not seen that. We are no longer using pynvml to detect GPU.

jperez999:
@oliverholworthy in terms of the edge case you outlined https://github.com/NVIDIA-Merlin/core/blob/ec9a360485bc48e67bb6fe4b5c5ca092eca2a97c/merlin/io/dataset.py#L249-L261 We would first check to see if HAS GPU was available and if it was, then we would check on cudf. This is the exact same as what happens on compat. We first decide if you have GPUs then we try to import the package. I dont see how this is an edge case. The only thing we could change is the error message since we now rely on nvml in dask instead of pynvml.

oliverholworthy:
> we cant ever use cudf without having a GPU, so that seems like the correct usage. Do you see a time when we can use cudf without a GPU?

Yeah that's true, aside from the segfault in earlier versions of cudf. you still can't use cudf without a gpu even it it's importable.

oliverholworthy:
I'm onboard with this change to wrap the imports in `compat` with `HAS_GPU`.

merlin/dtypes/mappings/cudf.py
```diff
@@ -41,17 +42,16 @@ def cudf_translator(raw_dtype) -> np.dtype:
...
+    try:
+        # We only want to register this mapping if cudf is available, even though
+        # the mapping itself doesn't use cudf (yet?)
 
-    import cudf  # pylint:disable=unused-import # noqa: F401
+        cudf_dtypes = DTypeMapping(
+            translator=NumpyPreprocessor("cudf", cudf_translator, attrs=["_categories"]),
+        )
+        _dtype_registry.register("cudf", cudf_dtypes)
+    except ImportError as exc:
```
oliverholworthy, line 54: do we still need this try/except here? what kind of `ImportError` can we get now that  we're checking `cudf`?

merlin/table/conversions.py
oliverholworthy: could changing this cause unexpected runtime errors later in the case of cudf being installed but no visible GPUs available?

tests/unit/io/test_avro.py
```diff
@@ -22,10 +22,12 @@
...
 import merlin.io
+from merlin.core.compat import cudf
 
-cudf = pytest.importorskip("cudf")
-dask_cudf = pytest.importorskip("dask_cudf")
-
+if cudf:
+    dask_cudf = pytest.importorskip("dask_cudf")
+else:
+    pytest.mark.skip(reason="cudf did not import successfully")
```
oliverholworthy, line 30: one intepretation of this is could suggest to someone reading this that it's only the cudf import that matters here. could be worth mentioning that we also need at lesast one visible CUDA device too?

PR #265, opened 2023-03-30
migrate configure tensorflow to core, separate has_gpu from compat

jperez999:
This PR moves the HAS_GPU flag into a separate file to avoid circular dependencies and it migrates the configure tensorflow method into core to be used with compat import of tensorflow.

### DIFF
--- a/merlin/core/compat.py
+++ b/merlin/core/compat.py
@@ -16,40 +16,17 @@
 
 # pylint: disable=unused-import
 import os
+import warnings
+
+from packaging import version
 
 try:
     from numba import cuda
 except ImportError:
     cuda = None
 
-from dask.distributed.diagnostics import nvml
-
-
-def _get_gpu_count():
-    """Get Number of GPU devices accounting for CUDA_VISIBLE_DEVICES environment variable"""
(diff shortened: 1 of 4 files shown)


merlin/core/compat.py
```diff
@@ -68,6 +130,69 @@ def _get_gpu_count():
...
+                    "If using TensorFlow < 2.3.0, you must install tfdlpack-gpu extension library"
+                )
+                raise ModuleNotFoundError(message) from e
+
+        else:
+            from tensorflow.experimental.dlpack import from_dlpack
+
+        return from_dlpack
+
+    configure_tensorflow()
```
oliverholworthy, line 194: would it be clearer to call this functoin in the libraries that need it instead of in the global scope?

oliverholworthy: this tensorflow import wasn't added in this PR, however it feels like a relevant place to discuss this. We seem to be following a path with this compat module to having a side effect of importing every framework module to use any core feature. For example to import the Dataset or Schema, we now have a side effect of importing and configuring tensorflow (if it's installed). Tensorflow is a good example of a package that is not  lightweight from an import perspective. I wonder how we might refacor this so that we can optionally import tensorflow only when required.

PR #270, opened 2023-04-03
compat imports update

jperez999:
This PR adds pandas and dask_cudf to compat imports.

### DIFF
--- a/merlin/core/compat.py
+++ b/merlin/core/compat.py
@@ -119,6 +119,11 @@ except ImportError:
     numpy = None
 
 try:
+    import pandas
+except ImportError:
+    pandas = None
+
+try:
     import cupy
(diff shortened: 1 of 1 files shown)


karlhigley (approved):

merlin/core/compat.py
```diff
@@ -118,6 +118,11 @@ def device_mem_size(kind="total", cpu=False):
 except ImportError:
     numpy = None
 
+try:
+    import pandas
```
oliverholworthy, line 122: where do we need to have a conditional import on pandas? Pandas is currently in the package requirements of core so should always be available.

karlhigley: I think this was just for consistency

PR #273, opened 2023-04-06
Return ragged shape when offests is provided to TensorColumn

oliverholworthy:
Part of https://github.com/NVIDIA-Merlin/Merlin/issues/255

This change updates the `TensorColumn` so that the shape and derived properties (`is_ragged`, for example) always correspond to a ragged representation when offsets is provided (is not None). 

Motivation for change is functionality in systems that depends on the `is_ragged` property. [In tensor_table_to_triton_request we check is_ragged on the column to decide if we should pass values/offsets in the inference request.](https://github.com/NVIDIA-Merlin/systems/blob/bf6cbc5bdb2302cdd8faeb05c8c8faa3126f57e7/merlin/systems/triton/conversions.py#L126)

### DIFF
--- a/merlin/table/tensor_column.py
+++ b/merlin/table/tensor_column.py
@@ -184,13 +184,8 @@ class TensorColumn:
     def _construct_shape(self, values, offsets):
         if offsets is not None:
             num_rows = len(offsets) - 1
-            row_lengths = offsets[1:] - offsets[:-1]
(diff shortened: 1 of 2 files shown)


oliverholworthy:
> figure out how to adjust the Systems code to check the right thing

For now systems only checks `is_ragged` on the tensor columns in two places. Here's a PR to check the representation using the offsets.

https://github.com/NVIDIA-Merlin/systems/pull/318

PR #274, opened 2023-04-06
fix cuda import logic from numba and device memsize

jperez999:
This PR fixes numba cuda import errors that allow import of cuda even when there are no cuda devices. Importing cuda is just importing the init file of the cuda submodule in numba but if you try to access cuda sub methods they fail. This fixes those issues by nullifying the import if cuda.is_available is false.

### DIFF
--- a/merlin/core/compat/__init__.py
+++ b/merlin/core/compat/__init__.py
@@ -91,7 +91,7 @@ def device_mem_size(kind="total", cpu=False):
             return psutil.virtual_memory().total
         elif kind == "free":
             return psutil.virtual_memory().free
-    elif cpu or not cuda:
+    elif cpu:
         warnings.warn("Please install psutil for full cpu=True support.")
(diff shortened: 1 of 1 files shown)


karlhigley (approved):

merlin/core/compat/__init__.py
```diff
@@ -17,16 +17,16 @@
...
 
 try:
     import psutil
 except ImportError:
     psutil = None
 
-try:
-    from numba import cuda
-except ImportError:
+if not cuda.is_available():
```
oliverholworthy, line 29: this might not be a problem with this function. However, I recall we had an issue once where we had `cuda.gpus.lst` in the global scope and that caused some problems with Dask-CUDA https://github.com/NVIDIA-Merlin/core/issues/144 It might be worth double checking this won't have a similar side effect?

PR #276, opened 2023-04-07
add hack to handle tf not recognizing bool dtype in dlpack

jperez999:
This PR adds a fix for tensorflow dlpack translation that does not allow bool dtypes. Because of this we have opted to cast all boolean columns as int8 dtype which is allowed by tensorflow.  Also adds newest cupy version to the requirements for GPU version of core, needed to ensure proper implementation of dlpack is used in tensortable. For reference:  https://github.com/tensorflow/tensorflow/blob/a9e6bd9ca4c0dbc96a71a1331bccf39a10757148/tensorflow/c/eager/dlpack.cc#L161-L242

### DIFF
--- a/merlin/table/cupy_column.py
+++ b/merlin/table/cupy_column.py
@@ -93,7 +93,9 @@ def _register_to_dlpack_from_cupy():
     import cupy as cp
 
     @_to_dlpack.register(cp.ndarray)
-    def _to_dlpack_from_tf_tensor(tensor):
+    def _to_dlpack_from_cp_tensor(tensor):
+        if tensor.dtype == cp.dtype("bool"):
+            tensor = tensor.astype(cp.dtype("int8"))
         return tensor
 
 
--- a/requirements-gpu.txt
+++ b/requirements-gpu.txt
@@ -3,4 +3,4 @@
 # cudf>=21.12
 # dask-cudf>=21.12
 # dask-cuda>=21.12
-# cupy>=7
+cupy-cuda11x
--- a/merlin/table/numpy_column.py
+++ b/merlin/table/numpy_column.py
@@ -111,4 +111,6 @@ def _register_from_numpy_to_dlpack_cpu():
 
     @_to_dlpack.register(np.ndarray)
     def _to_dlpack_cpu_from_numpy(array):
+        if array.dtype == np.dtype("bool"):
+            array = array.astype(np.dtype("int8"))
         return array
--- a/tox.ini
+++ b/tox.ini
@@ -29,6 +29,7 @@ setenv =
 deps =
     -rrequirements.txt
     -rrequirements-dev.txt
+    -rrequirements-gpu.txt
 commands =
     python -m pytest --cov-report term --cov merlin -rxs tests/unit
 


oliverholworthy (commented):
May be worth adding a test for this conversion to demonstrate support for converting bool types

merlin/table/cupy_column.py
```diff
@@ -93,7 +93,9 @@ def _register_to_dlpack_from_cupy():
     import cupy as cp
 
     @_to_dlpack.register(cp.ndarray)
-    def _to_dlpack_from_tf_tensor(tensor):
+    def _to_dlpack_from_cp_tensor(tensor):
+        if tensor.dtype == cp.dtype("bool"):
+            tensor = tensor.astype(cp.dtype("int8"))
```
oliverholworthy, line 98: Do we need a corresponding coercion back to bool dtypes on the other side (the from dlpack functions?)

jperez999: No we dont, we have seen in testing that the models seem to be ok with the conversion to 1 and 0 and they are able to interpret those values correctly. We were originally going to make that  change also (it belongs in the tensor column constructor) but during testing we saw that is was not necessary.

requirements-gpu.txt
```diff
@@ -3,4 +3,4 @@
 # cudf>=21.12
 # dask-cudf>=21.12
 # dask-cuda>=21.12
-# cupy>=7
+cupy-cuda11x
```
karlhigley, line 6: I think the version does matter here, since CuPy didn't add support for bools with DLpack until 12.0.0

merlin/table/cupy_column.py
karlhigley: I'm not 100% sure these values are being interpreted as bools on the other side (vs being treated as `int8` and having that still function for model training.) I'm not sure if it matters though, and I'd rather avoid an extra conversion if we can for performance reasons. So...I'm on the fence about adding the return conversion. "Let's see if it turns out to be a problem not to have it" seems reasonable, I suppose.

karlhigley (changes requested):
I think we need to specify a minimum dependency version for CuPy, and I'm not sure if we want to create a hard dependency on CUDA 11 since other toolkit versions also have CuPy 12.0.0 available. Seems like it might be better to specify the CuPy version we want as something like `cupy>=12.0.0`

requirements-gpu.txt
jperez999: We cant, when you try to specify that here it fails... When I try something like `pip install cupy-cuda11x>=12.0.0 it silently fails and nothing gets installed. This is after I have previously uninstalled the version in the environment. This is why I just left it to grab the latest version. 
image

jperez999: So you end up being forced to run the plain `pip install cupy-cuda11x` and you get the latest version.

karlhigley: I don't think we want to install `cupy-cuda11x` specifically though. In the containers we do, but here the requirement is actually to have `cupy>=12.0.0`—which we will by installing `cupy-cuda11x`, but would also be satisfied by `cupy-cuda12x`, `cupy-cuda102`, etc.

karlhigley: Okay, I'm wrong about that. `pip install cupy>=12.0.0` still tries to install from source even if you have `cupy-cuda*>=12.0.0`. Maybe we could make the various packages for satisfying the cupy dependency optional instead?

karlhigley (approved):

PR #279, opened 2023-04-12
Refactor `LocalExecutor` into more discrete steps that can be overridden

karlhigley:
This brings us closer to a point where we can get rid of the custom code for running NVTabular Workflows at serving time by aligning the structure of the code into distinct steps that can be used as is or overridden piecemeal as required.


oliverholworthy (approved):

oliverholworthy:
The refactor makes this a bit clearer than before. However, the steps that can be overridden part is perhaps not obvious. They _can_ be overridden. Whether they _should_ be is not so clear. Based on the names of the methods - implicitly private (from the `_` prefix naming convention). I suppose if this is a temporary workaround for our own use-case in systems it's ok.

PR #280, opened 2023-04-12
Feature/data format conversion

karlhigley:


merlin/dag/base_operator.py
```diff
@@ -355,6 +372,15 @@ def supports(self) -> Supports:
         """Returns what kind of data representation this operator supports"""
         return Supports.CPU_DATAFRAME | Supports.GPU_DATAFRAME
 
+    @property
+    def supported_formats(self) -> DataFormats:
```
oliverholworthy, line 376: This is saying that any new operator needs to support DataFrames and TensorTables as a default. I wonder if it would be simpler to have the default be only TensorTable? Since the implementation might need to be different for dataframes. And for Systems operators TensorTable is likely going to be the default?

karlhigley: Figuring a good default here is tricky, because the ops in NVT only support dataframes and the ops in Systems and the dataloaders only support `TensorTable`s. It didn't seem right to say that the default is to support zero formats, and at least for some simple cases dataframes and `TensorTables` are quasi-interchangeable, so this seemed like it kind of made sense. In practice though, the base operator classes in both NVT and Systems immediately change the default for all ops that inherit from them to just dataframes and just tensor tables (respectively.)

oliverholworthy (approved):

PR #284, opened 2023-04-13
Add support for from_dlpack with numpy 1.23.0

oliverholworthy:
Add support for from_dlpack with numpy 1.23.0

Check `np.from_dlpack` (added in `1.22.0`, renamed from private `_from_dlpack`), while keeping check of `np._from_dlpack` (added in 1.22.0) for backwards compatibility with numpy 1.22.0

### DIFF
--- a/merlin/table/numpy_column.py
+++ b/merlin/table/numpy_column.py
@@ -97,7 +97,13 @@ def _register_from_dlpack_cpu_to_numpy():
     @_from_dlpack_cpu.register(np.ndarray)
     def _from_dlpack_cpu_to_numpy(to, array):
         try:
+            # private `_from_dlpack` method added in 1.22.0
             return np._from_dlpack(array)
+        except AttributeError:
+            pass
+        try:
+            # public `from_dlpack` method added in 1.23.0
+            return np.from_dlpack(array)
         except AttributeError as exc:
             raise NotImplementedError(
                 "NumPy does not implement the DLPack Standard until version 1.22.0, "


merlin/table/numpy_column.py
```diff
@@ -97,7 +97,13 @@ def _register_from_dlpack_cpu_to_numpy():
     @_from_dlpack_cpu.register(np.ndarray)
     def _from_dlpack_cpu_to_numpy(to, array):
         try:
+            # private `_from_dlpack` method added in 1.22.0
             return np._from_dlpack(array)
```
karlhigley, line 101: I have a vague recollection that the semantics of these methods are actually not quite the same. Do we have a way to test against different versions of NumPy that would allow us to validate that both of these methods work as expected?

oliverholworthy: > np._from_dlpack(obj) function, where obj supports __dlpack__(), and returns an ndarray
> [NumPy 1.22.0 Release Notes](https://numpy.org/doc/1.23/release/1.22.0-notes.html?highlight=_from_dlpack#add-nep-47-compatible-dlpack-support)

> numpy.from_dlpack has been added to allow easy exchange of data using the DLPack protocol. It accepts Python objects that implement the __dlpack__ and __dlpack_device__ methods and returns a ndarray object which is generally the view of the data of the input object.
> [NumPy 1.23.0 Release Notes](https://numpy.org/doc/1.23/release/1.23.0-notes.html?highlight=from_dlpack#new-functions)

from the release notes the first private function in 1.22 mentions it supports objects that implement `__dlpack__`, while the notes of 1.23 mention support for objects with `__dlpack__` and `__dlpack_device__`. There was a bugfix for `__dlpack_device__` in https://github.com/numpy/numpy/pull/21138. Which maybe indicates we shouldn't be using this until we have a min version of numpy 1.23 ?

oliverholworthy: > Do we have a way to test against different versions of NumPy that would allow us to validate that both of these methods work as expected?

For automated tests we could potentially add a new workflow that has a test matrix for supported numpy versions? Combined with some pytest marks to run a subset of the tests? (the core tests run pretty quickly so that might not be strictly needed here)

karlhigley: Ah yeah, that's the difference. I don't think we do much with `__dlpack_device__` since it matters less on CPU. This is probably fine then.

karlhigley (approved):

PR #287, opened 2023-04-17
Return a dataframe type that matches `reader` passed to `fetch_table_data`

oliverholworthy:
Return a DataFrame type that matches `reader` passed to `fetch_table_data`.

This function is used in the `Categorify` operator in NVTabular and it's expected that the return value when passing a non-empty `table_cache` is to return a DataFrame with a type that matches the `reader` passed in.

## Example

```python
import pyarrow as pa
from merlin.io.worker import fetch_table_data

path = "path_to_data"
table_cache = {path: pa.Table.from_pydict({"feature": [1, 2]})}

fetch_table_data(table_cache, path, reader=pd.read_parquet)
# Before => cudf.DataFrame (if cudf available)
# After  => pd.DataFrame

### DIFF
--- a/merlin/io/worker.py
+++ b/merlin/io/worker.py
@@ -68,6 +68,7 @@ def fetch_table_data(
     """
     _lib = cudf if cudf else pd
     reader = reader or _lib.read_parquet
+    load_cudf = cudf and reader == cudf.read_parquet
     table = table_cache.get(path, None)
     cache_df = cache == "device"
     if table is None:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by karlhigley — APPROVED** [review]

File: merlin/io/worker.py [hunk 1]
```diff
@@ -60,55 +62,80 @@ def _get_worker_cache(name):
...
+                table_cache[path] = df.to_arrow()
+            elif isinstance(df, pd.DataFrame):
+                table_cache[path] = pa.Table.from_pandas(df)
                 cache_df = True
         if cats_only:
-            table.index.name = "labels"
-            table.reset_index(drop=False, inplace=True)
+            df.index.name = "labels"
+            df.reset_index(drop=False, inplace=True)
         if cache_df:
```
oliverholworthy, line 121: @karlhigley or @jperez999 do you recall why we sometimes save a dataframe and sometimes a pyarrow table in the cache? would there be any downside to changing this to always use one, the pyarrow table type for example?

PR #290, opened 2023-04-17
Add `actionlint` to pre-commit-config to check for valid GitHub Workflow config

oliverholworthy:
Add `actionlint` to pre-commit-config to check for valid GitHub Workflow config.

This makes it easier to spot if a workflow has not been picked up due to some invalid configuration or formatting.


oliverholworthy:
I'll update this to be workflow instead of pre-commit integration due to the concern about running pre-commit inside docker

karlhigley (approved):

PR #296, opened 2023-04-27
Make `ColumnSelector.all` a property instead of a manually set attribute

karlhigley:
This seems to be causing situations where `ColumnSelector.all` doesn't exist, though I'm not entirely sure how. In any case, making it a property is a better design anyway, so let's see if this fixes the issues.

### DIFF
--- a/tests/unit/dag/test_column_selector.py
+++ b/tests/unit/dag/test_column_selector.py
@@ -270,6 +270,12 @@ def test_applying_inverse_selector_to_schema_selects_relevant_columns():
     assert result == schema
 
 
+def test_wildcard_selector_bool():
+    selector = ColumnSelector("*")
+
+    assert selector.all is True
+
+
 def test_wildcard_selection():
     selector = ColumnSelector("*")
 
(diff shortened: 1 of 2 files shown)


tests/unit/dag/test_column_selector.py
```diff
@@ -270,6 +270,12 @@ def test_applying_inverse_selector_to_schema_selects_relevant_columns():
     assert result == schema
 
 
+def test_wildcard_selector_bool():
+    selector = ColumnSelector("*")
+
+    assert selector.all is True
```
oliverholworthy, line 276: It looks like this test would have passed without this change. I'm not sure if this PR has a functional change, though the property method is clearer in terms of documenting the public interface.

oliverholworthy (approved):

tests/unit/dag/test_column_selector.py
karlhigley: I'm right there with you on thinking this shouldn't be a functional change, but if you see the linked bug ticket, somehow `self.all` is not getting set/created on selector objects. I think this should address that issue, although I'm not entirely sure why the code wasn't working as written.

PR #299, opened 2023-04-28
Adjust the `DaskExecutor` API methods to take `Dataset`s instead of ddfs

karlhigley:
This lays the groundwork for creating an operator that runs a DAG sub-graph, which will require being able to `fit()` the subgraph in NVT.

Required by https://github.com/NVIDIA-Merlin/NVTabular/pull/1807

### DIFF
--- a/merlin/dag/executors.py
+++ b/merlin/dag/executors.py
@@ -333,12 +333,12 @@ class DaskExecutor:
         )
 
     def fit(self, dataset: Dataset, graph, strict=False):
-        """Calculates statistics for a set of nodes on the input dataframe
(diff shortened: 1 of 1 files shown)


merlin/dag/executors.py
```diff
@@ -274,23 +280,12 @@ def transform(
         Transforms all partitions of a Dask Dataframe by applying the operators
```
nv-alaiacano, line 280: dont' forget to update the docstring(s) from "Dask Dataframe" to "Dataset"

nv-alaiacano (approved):

PR #300, opened 2023-04-28
Update packages workflow, separating PyPI from conda build

oliverholworthy:
Update packages workflow, separating PyPI from conda build

### DIFF
--- a/.github/workflows/packages.yaml
+++ b/.github/workflows/packages.yaml
@@ -1,37 +1,29 @@
-name: Build Core Packages (CPU)
+name: Packages
 
 on:
   workflow_dispatch:
   push:
-    branches: [main]
+    branches:
+      - main
+      - release*
     tags:
       - v*
   pull_request:
     branches: [main]
 
 jobs:
-  build-packages:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        python-version: [3.8]
-        os: [ubuntu-latest]
-
+  build-pypi:
+    name: Build PyPI Package
+    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/packages.yaml
```diff
@@ -40,55 +32,45 @@ jobs:
...
+          echo "conda_package : $conda_package"
+          echo "conda_package=$conda_package" >> "$GITHUB_OUTPUT"
       - name: Upload conda artifacts to github
         uses: actions/upload-artifact@v3
         with:
           name: conda
           path: ${{ steps.conda_build.outputs.conda_package }}
 
-      # Build docs, treat warnings as errors
-      - name: Building docs
```
oliverholworthy, line 63: This Building docs job has recently started failing with Python 3.8. Removing this, since it doesn't seem relevant to the package build/release

karlhigley: I think it may have been included here to make sure we don't release when the docs build is broken?

oliverholworthy: I've added that back as a required job that runs before the package build/release jobs. And removed the dependency  between this Packages workflow and the PR docs Preview Workflow by moving the docs build job with the artifact upload closer to where it's used.

karlhigley (approved):

.github/workflows/packages.yaml
oliverholworthy: To get the Docs PR Preview running as an independent workflow (not triggered by the successful completion of another workflow requires some further work on that re-usable workflow). 

I updated this PR to be only related to the splitting of the jobs into PyPI/Conda build/release only since that is an independent issue

PR #301, opened 2023-04-28
Create a `Subgraph` operator and integrate it with `Graph`/`Executors`

karlhigley:
This provides a way to run one `Graph` inside another `Graph`, which is useful for a bunch of things:
* Separating ETL code (for standardizing offline logs with online request data) from feature transformation code (run on both training/eval data and online requests) 
* Separating user and item sub-graphs inside NVT workflows so they can be run separately at serving time
* Running the same sub-graph multiple times to implement iterative processes (like calibration when building a list of recs)

Depends on https://github.com/NVIDIA-Merlin/core/pull/299

### DIFF
--- a/merlin/dag/executors.py
+++ b/merlin/dag/executors.py
@@ -14,8 +14,10 @@
 # limitations under the License.
 #
 import logging
+from typing import Union
 
 import dask
+import dask.dataframe as dd
 import pandas as pd
 from dask.core import flatten
 
@@ -27,9 +29,11 @@ from merlin.core.utils import (
     set_client_deprecated,
 )
 from merlin.dag import ColumnSelector, Graph, Node
+from merlin.dag.stat_operator import StatOperator
 from merlin.dtypes.shape import DefaultShapes
 from merlin.io import Dataset
 from merlin.io.worker import clean_worker_cache
+from merlin.schema import Schema
 
 LOG = logging.getLogger("merlin")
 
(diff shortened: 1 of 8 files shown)


merlin/dag/executors.py
```diff
@@ -332,8 +334,64 @@ def transform(
...
+                dependencies.difference_update(current_phase)
+
+        # This captures the output dtypes of operators like LambdaOp where
+        # the dtype can't be determined without running the transform
+        self._transform_impl(dataset, graph, capture_dtypes=True).sample_dtypes()
+        graph.construct_schema(schema, preserve_dtypes=True)
+
+        return graph
+
+    def _fit_ops(self, dataset: Dataset, graph, strict=False):
```
karlhigley, line 393: This part of the code was already in Core and was called by the code above that moved from NVT

merlin/dag/executors.py
```diff
@@ -332,8 +334,64 @@ def transform(
             )
         )
 
-    def fit(self, dataset: Dataset, graph, strict=False):
-        """Calculates statistics for a set of nodes on the input dataset
+    def fit(self, dataset: Dataset, graph: Graph, schema: Schema = None) -> "Graph":
```
karlhigley, line 337: Fit was in a weird state where half the code was in NVT and the other half was here in Core. Fitting currently really only exists in NVT, so that wasn't problematic until we needed to test fitting on functionality that only exists in Core. So, this code moved (largely as is) from NVT to Core

merlin/dag/graph.py
```diff
@@ -36,23 +36,22 @@ class Graph:
...
         parents_with_deps.append(output_node)
 
-        for name, sg in self.subgraphs.items():
-            if sg not in parents_with_deps:
-                raise ValueError(
-                    f"The output node of subgraph {name} does not exist in the provided graph."
-                )
+        for node in postorder_iter_nodes(output_node):
+            op = node.op
+            if op.is_subgraph and op.name:
```
karlhigley, line 48: The `is_subgraph` property is a workaround to avoid the circular import issue that would result from making `Graph` depend on `Subgraph` in order to do an `isinstance` check on this line

merlin/dag/graph.py
```diff
@@ -198,6 +197,16 @@ def remove_inputs(self, to_remove):
     def get_nodes_by_op_type(cls, nodes, op_type):
         return set(node for node in iter_nodes(nodes) if isinstance(node.op, op_type))
 
+    def clear_stats(self):
```
karlhigley, line 200: Moved from NVT

merlin/dag/ops/subgraph.py
```diff
@@ -0,0 +1,148 @@
...
+        """Removes calculated statistics from each node in the graph
+
+        See Also
+        --------
+        StatOperator.clear
+        """
+        for stat in Graph.get_nodes_by_op_type([self.graph.output_node], StatOperator):
+            stat.op.clear()
+
+    def column_mapping(self, col_selector):
```
karlhigley, line 140: It turns out to be crucially important that the `Subgraph` operator properly delegates the job of creating a mapping that connects outputs with inputs 😅

merlin/dag/stat_operator.py
```diff
@@ -0,0 +1,55 @@
+#
```
karlhigley, line 1: Moved from NVT in its entirety

merlin/dag/executors.py
```diff
@@ -379,6 +437,24 @@ def fit(self, dataset: Dataset, graph, strict=False):
         for computed_stats, node in zip(results, nodes):
             node.op.fit_finalize(computed_stats)
 
+    def _transform_impl(self, dataset: Dataset, graph: Graph, capture_dtypes=False):
```
karlhigley, line 440: This came from NVT, where the name `transform_impl` made more sense. Debating what a better name for this would be (or if it should just be inlined.)

oliverholworthy: Looks like we may not need most of this method since the `self.transform` now always returns a Dataset and the graph construction is already applied in the `fit` method were it's called. So inlining might only require the line that calls `self.transform`?

merlin/dag/ops/subgraph.py
```diff
@@ -0,0 +1,148 @@
...
+            Returns a transformed dataframe for this operator
+        """
+        executor = LocalExecutor()
+        transformable = executor.transform(transformable, self.graph)
+        while self.loop_until and not self.loop_until(transformable):
+            transformable = executor.transform(transformable, self.graph)
+        return transformable
+
+    def fit(self, col_selector: ColumnSelector, dataset: Dataset):
+        DaskExecutor().fit(dataset, self.graph)
```
oliverholworthy, line 72: It feels a bit unexpected that the `transform` uses `LocalExecutor`, while the `fit` uses `DaskExecutor`. Maybe don't need to action this in this PR, but something doesn't feel right about that. Are we thinking of extending support for `fit` beyond the dask versions? The addition of the StatOperator in this PR suggests we might be?

oliverholworthy:
I merged in the main branch that had some fixes for some unrelated package build checks. We've now got a few failing tests about `unexpected keyword argument 'subgraphs'`.

PR #312, opened 2023-05-04
Add `Rename` to the set of core DAG ops that work in all DAGs

karlhigley:
This should work with either dataframes or `TensorTables`, which requires minor changes to TT to add a definition of equality and make it possible to change column names.

Depends on #318

### DIFF
--- a/merlin/table/tensor_table.py
+++ b/merlin/table/tensor_table.py
@@ -209,6 +209,12 @@ class TensorTable:
         """
         return list(self.keys())
 
+    @columns.setter
+    def columns(self, col_names):
+        for col, col_name in zip(self.columns, col_names):
+            self._columns[col_name] = self._columns[col]
+            del self._columns[col]
+
     @property
     def column_type(self):
         return type(list(self.values())[0])
--- a/merlin/table/tensor_table.py
+++ b/merlin/table/tensor_table.py
@@ -247,10 +247,26 @@ class TensorTable:
         return result
 
     def cpu(self):
+        """
+        Move this TensorTable and its columns to CPU
+
+        Returns
+        -------
+        TensorTable
+            A new TensorTable containing the same columns but on CPU
+        """
         columns = {col_name: col_values.cpu() for col_name, col_values in self.items()}
         return TensorTable(columns)
 
     def gpu(self):
+        """
+        Move this TensorTable and its columns to GPU
+
+        Returns
+        -------
+        TensorTable
+            A new TensorTable containing the same columns but on GPU
+        """
         columns = {col_name: col_values.gpu() for col_name, col_values in self.items()}
         return TensorTable(columns)
 
--- a/merlin/table/tensor_table.py
+++ b/merlin/table/tensor_table.py
@@ -211,9 +211,10 @@ class TensorTable:
 
     @columns.setter
     def columns(self, col_names):
+        renamed_columns = {}
         for col, col_name in zip(self.columns, col_names):
-            self._columns[col_name] = self._columns[col]
-            del self._columns[col]
+            renamed_columns[col_name] = self._columns[col]
+        self._columns = renamed_columns
 
     @property
     def column_type(self):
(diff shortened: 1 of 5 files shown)


merlin/table/tensor_table.py
```diff
@@ -145,6 +155,12 @@ def columns(self):
         """
         return list(self.keys())
 
+    @columns.setter
+    def columns(self, col_names):
+        for col, col_name in zip(self.columns, col_names):
+            self._columns[col_name] = self._columns[col]
+            del self._columns[col]
```
oliverholworthy, line 162: there could be an edge case here where a column is being renamed to an already existing column, which might result in missing columns in the result? e.g. one simple test case might be to swap labels between two columns

merlin/core/protocols.py
```diff
@@ -91,9 +91,14 @@ class Transformable(DictLike, Protocol):
     base to do without them.
     """
 
+    @property
```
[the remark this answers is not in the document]
oliverholworthy: one option could be to use `singledispatchmethod` like we did in the NVTabular Workflow transform method in https://github.com/NVIDIA-Merlin/NVTabular/pull/1777

tests/unit/dag/ops/test_rename.py
```diff
@@ -0,0 +1,58 @@
...
+
+    op = Rename(name="z")
+    transformed = op.transform(selector, df)
+    expected = transformable({"z": np.array([1, 2, 3, 4, 5])})
+    assert all(transformed == expected)
+
+    op = Rename(f=lambda name: name.upper())
+    transformed = op.transform(selector, df)
+    expected = transformable({"X": np.array([1, 2, 3, 4, 5])})
+    assert all(transformed == expected)
```
oliverholworthy, line 58: This is will return True for DataFrames as long as the columns are different because the return type of `DataFrame == DataFrame` is a DataFrame, and `all(DataFrame)` iterates through the column names and asserts they are all truthy values. [DataFrame.eq](https://pandas.pydata.org/pandas-docs/version/1.5/reference/api/pandas.DataFrame.eq.html) performs element-wise equality.

If we mostly need to know if a DataFrame contains the same elements and don't need to operate on the element-wise equality result, maybe could use [DataFrame.equals](https://pandas.pydata.org/pandas-docs/version/1.5/reference/api/pandas.DataFrame.equals.html) or [pd.testing.assert_frame_equal](https://pandas.pydata.org/pandas-docs/version/1.5/reference/api/pandas.testing.assert_frame_equal.html#pandas.testing.assert_frame_equal) to assert dataframe equality instead.

If we implemented an `equals` method on TensorTable instead of element-wise `__eq__` we might not immediately need the `__eq__` for this use-case. However, if we do add that to `TensorTable` and would like to be consistent with the Series/DataFrame `== / __eq__` equivalent perhaps we would return a TensorTable from this with element-wise comparison At the moment it looks like we're returning a tensor type. 

If the motivation is about being able to write unit tests that transform both TensorTable and DataFrame inputs. Having a helper funciton in core could be another option `assert_frame_or_table_equal`.

oliverholworthy: One benefit of `assert_frame_equal` over `DataFrame.equals` I can think of are error messages that explain why the dataframe doesn't match. `equals` returns True or False always and doesn't intend to raise exceptions or print messages, while `assert_frame_equal` is designed for testing and will explain why the dataframe doesn't match.


```python
pd.testing.assert_frame_equal(pd.DataFrame({"a": [1.0, 2.0]}), pd.DataFrame({"a": [1, 2]}))
# =>
# AssertionError: Attributes of DataFrame.iloc[:, 0] (column name="a") are different
# 
# Attribute "dtype" are different
# [left]:  float64
# [right]: int64


pd.DataFrame({"a": [1.0, 2.0]}).equals(pd.DataFrame({"a": [1, 2]}))
# => False
```

Another feature is by default it will use an tolerance check to avoid small floating point differences

karlhigley: This PR has been updated to add an `assert_transformable_equal` testing helper and to remove changes to `TensorTable` and `TensorColumn`'s equality definitions. Those will still need some changes to standardize them with DataFrames and Series, but those changes aren't required to get `Rename` moved into Core.

oliverholworthy (approved):

PR #314, opened 2023-05-08
Fix Pandas extension dtype mapping for newer versions of Pandas

karlhigley:
Newer versions of Pandas seem to have changed `.numpy_dtype` from a property to a method, which has broken how Pandas extension dtypes are mapped back to Meriln dtypes (via NumPy as an intermediary.) This fix adjusts for that, and for recent changes that allowed returning `mn.unknown` instead of raising an error for unknown types.

### DIFF
--- a/merlin/dtypes/__init__.py
+++ b/merlin/dtypes/__init__.py
@@ -16,6 +16,7 @@
 
 # flake8: noqa
 from merlin.dtypes import mappings
+from merlin.dtypes import aliases
 from merlin.dtypes.aliases import *
 from merlin.dtypes.base import DType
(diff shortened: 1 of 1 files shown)


oliverholworthy:
Are there any tests that currently fail prior to this PR? If not, maybe we need an extra test to demonstrate this?

oliverholworthy (approved):

karlhigley:
Merging this to unblock builds, but on the issue of tests, this change allows tests in NVT's `test_dask_nvt.py` to pass so it is covered by tests but unfortunately not in this repo. I think adding a test here is a good idea, and I will try to put together a PR that replicates the issue with a simpler test than the ones in NVT.

PR #318, opened 2023-05-11
Adjust the definition of `TensorTable` equality to match `DataFrames`

karlhigley:
It's difficult to write operators and test them without similar definitions of equality for `DataFrames` and `Series` vs `TensorTables` and `TensorColumns`. This PR adjusts them to be the same, so that we can parametrize tests across `DataFrame` and `TensorTable` types in order to make sure that operator transforms work interchangeably on both formats.

### DIFF
--- a/merlin/table/tensor_column.py
+++ b/merlin/table/tensor_column.py
@@ -66,7 +66,13 @@ class TensorColumn:
             return object.__new__(cls)
 
     def __init__(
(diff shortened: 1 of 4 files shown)


karlhigley:
Closing this, because @oliverholworthy has pointed out some significant issues with these definitions in terms of how they deviate from dataframes. We should still do this, but we can do better than this PR.

PR #324, opened 2023-05-22
Migrate fit, stat operator, lambda op from nvtabular

jperez999:
This PR migrates certain key components from nvtabular to core so that they can be reused by more parts of merlin. 

Migrated Fit to executors in core from nvtabular workflow
Migrated stat operator to core
migrated lambda op to core, and renamed to UDF (user defined function)


merlin/dag/executors.py
```diff
@@ -372,7 +374,76 @@ def transform(
...
+            self.transform(ddf, graph.output_node, graph.output_dtypes, capture_dtypes=True),
+            cpu=dataset.cpu,
+            base_dataset=dataset.base_dataset,
+            schema=graph.output_schema,
+        ).sample_dtypes()
+        graph.construct_schema(dataset.schema, preserve_dtypes=True)
+
+        return graph
+
+    def fit_phase(self, ddf, nodes, strict=False):
```
karlhigley, line 446: 👍🏻 👍🏻 👍🏻

PR #331, opened 2023-05-23
Use compat.cupy to determine whether to run tests on GPU in test_udf

oliverholworthy:
Use `compat.cupy` to determine whether to run tests on GPU in test_udf. Fixes tests in GPU environment where devices are not visible. Follow up to #324


karlhigley (approved):
Thanks for this!

PR #333, opened 2023-05-23
added fitted attr to stat op and checks in executor

jperez999:
This PR adds the fitted (bool) attribute to all Stat Operators. This boolean is set to true after an operator has been fit. You can now manually specify to "refit" a fitted StatOperator in the fit method call. And transform checks to ensure all Stat Operators have been fitted before running transform on a graph.

### DIFF
--- a/merlin/dag/ops/stat_operator.py
+++ b/merlin/dag/ops/stat_operator.py
@@ -27,6 +27,8 @@ class StatOperator(BaseOperator):
     on top of the Operator class.
     """
 
+    fitted = False
+
     def fit(self, col_selector: ColumnSelector, ddf: dd.DataFrame) -> Any:
         """Calculate statistics for this operator, and return a dask future
         to these statistics, which will be computed by the workflow."""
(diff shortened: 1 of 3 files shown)


merlin/dag/ops/stat_operator.py
```diff
@@ -27,6 +27,8 @@ class StatOperator(BaseOperator):
     on top of the Operator class.
     """
 
+    fitted = False
```
karlhigley, line 30: I'm still somewhat concerned that this variable needs to be on the instance level instead of the class level. It works in our test case because we only have one op that needs fitting, but I have a hunch this is going to cause weird errors that are difficult to troubleshoot when we have more than one.

oliverholworthy (approved):

merlin/dag/ops/stat_operator.py
oliverholworthy: Unless we have some code that runs `StatOperator.fitted = True`, changing the class-level attribute we should be ok? If we're concerned about that, maybe can implement an `__init__` method that sets the fitted attribute?

PR #337, opened 2023-05-25
add npy export capability to dataset

jperez999:
This PR adds support for exporting a dataset to an npy file. There are certain caveats and checks that must abided by in order for this to work. It allows for two methods of writing, first as one large dataframe dump, second is with append, meaning you can have larger than memory sized data added to the file. List Columns must be exploded via the column axis to write to an npy file. A helper method has been created to facilitate that action.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,6 +11,7 @@ tqdm>=4.0
 tensorflow-metadata>=1.2.0
 betterproto<2.0.0
 packaging
+npy_append_array
 
 # pynvml==11.5.0 is incompatible with distributed<2023.2.1
 pynvml>=11.0.0,<11.5
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ tqdm>=4.0
 tensorflow-metadata>=1.2.0
 betterproto<2.0.0
 packaging
-npy_append_array
+npy-append-array
 
 # pynvml==11.5.0 is incompatible with distributed<2023.2.1
 pynvml>=11.0.0,<11.5
--- a/tests/unit/io/test_dataset.py
+++ b/tests/unit/io/test_dataset.py
@@ -13,11 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import glob
+
+import numpy as np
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -11,6 +11,7 @@ tqdm>=4.0
 tensorflow-metadata>=1.2.0
 betterproto<2.0.0
 packaging
+npy-append-array
```
karlhigley, line 14: Not sure we want to add this as a hard requirement for Core, which would make it a hard requirement for all of Merlin. What approaches do we have to make it optional somehow?

jperez999: I made it a hard requirement because it is now a core capability of the dataset class. I think all APIs for the dataset are available in all environments. But if we want to make it optional we could break it out into a separate requirements file and have the ability to install additionally like merlin-core[np-append]...?

karlhigley: It looks like the append array functionality has been merged into the mainline of NumPy as of this PR: https://github.com/numpy/numpy/pull/20321

karlhigley: Resolution here is to ahead and make the package a hard dependency, since the package's functionality will be part of the NumPy 1.25 release, which alleviates concerns about ongoing maintenance.

tests/unit/io/test_dataset.py
```diff
@@ -55,3 +58,49 @@ def test_infer_list_dtype_unknown():
...
+            "id": [1, 2, 3, 4, 5, 6],
+            "embed_1": [1, 2, 3, 4, 5, 6],
+            "embed_2": [1, 2, 3, 4, 5, 6],
+            "embed_3": [1, 2, 3, 4, 5, 6],
+        }
+    )
+    dataset = Dataset(df)
+    path = str(tmpdir / "result.npy")
+    dataset.to_npy(path, append=append)
+    nparr = np.load(path, allow_pickle=True)
```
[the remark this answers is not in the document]
jperez999: I will update though, because it is only required when you have an object so the last two test cases do not actually need that allow-pickle option. It was an artifact of copying and pasting the original test (then changing the data).

PR #338, opened 2023-05-30
adding rapids runner

jperez999:
testing activation of rapids runner.

### DIFF
--- a/.github/workflows/gpu-ci.yml
+++ b/.github/workflows/gpu-ci.yml
@@ -3,16 +3,14 @@ name: GPU CI
 on:
   workflow_dispatch:
   push:
-    branches: [main]
-    tags:
-      - "v[0-9]+.[0-9]+.[0-9]+"
+    branches-ignore: [main]
   pull_request:
     branches: [main]
     types: [opened, synchronize, reopened]
 
 jobs:
   gpu-ci:
-    runs-on: 1GPU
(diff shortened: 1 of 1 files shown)


.github/workflows/gpu-ci-docker.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Get Branch name
+        id: get-branch-name
+        uses: NVIDIA-Merlin/.github/actions/branch-name@6f0539fba24f60da2aee63c5925bee7cee3206e3
+      - name: Run tests
+        run: |
+          merlin_branch="${{ steps.get-branch-name.outputs.branch }}"
+          MERLIN_BRANCH=$merlin_branch COMPARE_BRANCH=$merlin_branch \
```
oliverholworthy, line 29: These env vars are not required here. 

- `MERLIN_BRANCH` is used in other places where we need to install other merlin packages from the equivalent branch ( release branch or development branch). In this `test-gpu` tox environment we only install the core package.
  - the `get-branch-name` step can also be removed here too since we won't use it if `MERLIN_BRANCH` is removed
- `COMPARE_BRANCH` is something currently only used in Merlin Models to inspect the diff and label parts as changed or unchanged (so that tests using TensorFlow only run if the tensorflow code has changed).

PR #342, opened 2023-06-08
Fix udf op

jperez999:
This PR fixes issues with the UDF op outlined here https://github.com/NVIDIA-Merlin/systems/issues/360. This issue occurs when the user defined function is created with a targeted framework. However if the function is setup with a specific framework, the output would end up as the framework specified. This causes issues with the expected output (which should be the same as the input). The fix added here uses a dictionary instead of an input type collection. After the collection is complete we create a dataframe to ensure continued support for downstream operators. This is necessary to be able to all the different types of inputs possible (i.e. TensorTable, Cudf Dataframe, Pandas Dataframe). If using tensortable, it will be converted before the next operator in the executor.

### DIFF
--- a/merlin/dag/ops/udf.py
+++ b/merlin/dag/ops/udf.py
@@ -15,6 +15,7 @@
 #
 from inspect import getsourcelines, signature
 
+from merlin.core.dispatch import make_df
 from merlin.core.protocols import Transformable
 from merlin.dag.base_operator import BaseOperator
 from merlin.dag.selector import ColumnSelector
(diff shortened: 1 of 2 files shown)


merlin/dag/ops/udf.py
```diff
@@ -78,7 +79,8 @@ def transform(
             else:
                 # shouldn't ever happen,
                 raise RuntimeError(f"unhandled UDF param count {self._param_count}")
-        return new_df
+        # return input type data
+        return make_df(new_df)
```
oliverholworthy, line 83: This works ok for data frame based  graphs, but we might need to revisit in a follow-up change to make this support tensortable too?

oliverholworthy:
It's probably worth adding a test for this situation where we pass a pandas data frame but the user defined function returns a cudf series and/or the other way around

merlin/dag/ops/udf.py
jperez999: No, I think its fine, for tensortable it also works because the executor handles the change to something compatible for the next operator. https://github.com/NVIDIA-Merlin/core/blob/main/merlin/dag/executors.py#L105

jperez999:
> It's probably worth adding a test for this situation where we pass a pandas data frame but the user defined function returns a cudf series and/or the other way around

Test added.

oliverholworthy (approved):

merlin/dag/ops/udf.py
oliverholworthy: what about if a user-defined function returns a TensorColumn type, is that something we'd like to support?

PR #343, opened 2023-06-09
changes required to make subgraphs work in nvt

jperez999:
This PR has some changes that fix some issues with graphs and it refactors the iter_nodes method in node to handle flattening of nodes in graphs that are subgraphs. These changes are required to make subgraphs work in nvtabular.

### DIFF
--- a/merlin/dag/executors.py
+++ b/merlin/dag/executors.py
@@ -512,7 +512,7 @@ class DaskExecutor:
 
         dask_client = global_dask_client()
         if dask_client:
-            results = [r.result() for r in dask_client.compute(stats)]
+            results = [r.result() for r in dask_client.compute(stats) if r is not None]
         else:
             results = dask.compute(stats, scheduler="synchronous")[0]
         for computed_stats, node in zip(results, nodes):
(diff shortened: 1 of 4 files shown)


merlin/dag/executors.py
```diff
@@ -512,7 +512,7 @@ def fit_phase(self, dataset, nodes, strict=False):
 
         dask_client = global_dask_client()
         if dask_client:
-            results = [r.result() for r in dask_client.compute(stats)]
+            results = [r.result() for r in dask_client.compute(stats) if r is not None]
```
oliverholworthy, line 515: what's the scenario where the fit method of a on (Subraph op) returns None?

jperez999: Always, because it does not have stats itself. It has operators within it that have stats. So it will return the stats when its graph runs the fit but it wont bubble that up.

oliverholworthy (approved):

merlin/dag/executors.py
oliverholworthy: > stats is empty and then there is not r

if stats is empty (`[]`) then the dask client compute also returns an empty list `dask_client.compute([]) => []`. and r is never reached as None in that case.

PR #344, opened 2023-06-09
migrate metadata operators to core

jperez999:
This PR migrates all the add metadata operators to core. This will allow them to be used in other parts of merlin, like systems. These are general operators that manipulate schemas. They are not required to be located in the nvtabular repo. By moving these operators to core, we remove the nvtabular import requirement for these ops.

### DIFF
--- a/merlin/dag/ops/add_metadata.py
+++ b/merlin/dag/ops/add_metadata.py
@@ -13,9 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from merlin.core.protocols import Transformable
 from merlin.dag.base_operator import BaseOperator
-from merlin.dag.selector import ColumnSelector
 from merlin.schema.tags import Tags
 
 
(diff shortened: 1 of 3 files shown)


merlin/dag/ops/add_metadata.py
```diff
@@ -0,0 +1,75 @@
...
+    This operator will add user defined tags and properties
+    to a Schema.
+    """
+
+    def __init__(self, tags=None, properties=None):
+        super().__init__()
+        self.tags = tags or []
+        self.properties = properties or {}
+
+    def transform(
```
oliverholworthy, line 33: This appears to be the default implementation of `transform` in the `BaseOperator`. Could this be omitted or, are we adding this here as part of a convention of adding a transform for every operator?

oliverholworthy:
Should we add these to `merlin/dag/ops/__init__.py` like the other ops we have in Core? so that they can be imported from `merlin.dag.ops` instead of requiring an import of this `add_metadata` module?

PR #349, opened 2023-06-21
add changes to support subgraph in multistage example

jperez999:
This PR updates base_operator, subgraph and graph. In base operator we add the artifact hooks (load and save). This allows for loading and saving of artifacts in the graph. The graphs adds the capability to load and save artifacts so that systems operators can run on ensembles that use a local executor. We also added the set_storage_path to the subgraph so that all nodes in a subgraph that need to update the location of anything that might be stored can also update the path in the operator.

### DIFF
--- a/merlin/dag/graph.py
+++ b/merlin/dag/graph.py
@@ -49,6 +49,8 @@ class Graph:
 
         self.subgraphs: Dict[str, "Graph"] = {}
         _find_subgraphs(output_node, self.subgraphs)
+        for node in list(postorder_iter_nodes(self.output_node, flatten_subgraphs=True)):
+            node.op.load_artifacts("")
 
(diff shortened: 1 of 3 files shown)


merlin/dag/graph.py
```diff
@@ -49,6 +49,8 @@ def __init__(self, output_node: Node):
 
         self.subgraphs: Dict[str, "Graph"] = {}
         _find_subgraphs(output_node, self.subgraphs)
+        for node in list(postorder_iter_nodes(self.output_node, flatten_subgraphs=True)):
+            node.op.load_artifacts("")
```
karlhigley, line 53: Does it still work if we omit the `""` arg here?

karlhigley (approved):

PR #350, opened 2023-06-22
Make output of transform same type as input

jperez999:
This PR makes it so that the output to an executor transform is the same type as the input.

### DIFF
--- a/merlin/dag/executors.py
+++ b/merlin/dag/executors.py
@@ -311,10 +311,6 @@ class DaskExecutor:
         if client is not None:
             set_client_deprecated(client, "DaskExecutor")
 
-    @property
-    def target_format(self):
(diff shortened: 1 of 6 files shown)


merlin/dag/executors.py
```diff
@@ -300,6 +311,10 @@ def __init__(self, client=None):
         if client is not None:
             set_client_deprecated(client, "DaskExecutor")
 
+    @property
+    def target_format(self):
+        return DataFormats.PANDAS_DATAFRAME | DataFormats.CUDF_DATAFRAME
+
```
karlhigley, line 317: I think this is a stray change that's not required yet (until a subsequent PR re: running Systems ops in the `DaskExecutor`.)

karlhigley (approved):

PR #353, opened 2023-06-29
fix the column merging during graph traversal and concat selector 

jperez999:
This PR fixes graph traversal column merging. This is to allow for retreival of root columns and support overlapping column usage. The new behavior is to return whatever the last node of the graph is requiring for a target column. We also fixed the concat column op selector computation logic. Parents + dependencies were overriding input schema columns unnecessarily. That is only required when propagating column groups.

### DIFF
--- a/tests/unit/dag/test_graph.py
+++ b/tests/unit/dag/test_graph.py
@@ -15,9 +15,12 @@
 #
 import pytest
 
+from merlin.core.dispatch import make_df
 from merlin.dag import Graph, Node, iter_nodes, postorder_iter_nodes, preorder_iter_nodes
 from merlin.dag.base_operator import BaseOperator
+from merlin.dag.executors import LocalExecutor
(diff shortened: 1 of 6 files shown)


tests/unit/dag/test_graph.py
```diff
@@ -95,3 +98,42 @@ def test_subgraph_with_summed_subgraphs():
     assert post_len == pre_len
     assert iter_len == post_len
     assert iter_len == pre_len
+
+
+def test_selector_concat_parents():
```
karlhigley, line 103: It looks like this tests that an overlapping column name prefers the version from the right-hand side. Not quite sure how to distill that into a good test name though...`test_concat_prefers_rhs`?

karlhigley (approved):

karlhigley:
It looks like this change breaks a few tests in NVTabular that probably actually are related to this change

PR #357, opened 2023-07-06
Inf node merge

jperez999:
This PR merges the inference operator and node with the base operator and node, respectively. This is part of an effort to make all operators work throughout merlin (workflow + ensemble).

### DIFF
--- a/merlin/dag/base_operator.py
+++ b/merlin/dag/base_operator.py
@@ -20,6 +20,7 @@ from enum import Flag, auto
 from typing import Any, List, Optional, Union
 
 import merlin.dag
+import merlin.dag.utils
 from merlin.core.protocols import Transformable
 from merlin.dag.selector import ColumnSelector
 from merlin.schema import ColumnSchema, Schema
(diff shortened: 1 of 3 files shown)


merlin/dag/base_operator.py
```diff
@@ -409,3 +410,92 @@ def _get_columns(self, df, selector):
...
+        """
+        raise NotImplementedError(
+            "Exporting an operator to run in a particular context (i.e. Triton) "
+            "only makes sense when a runtime is specified. To select an "
+            f"operator for the appropriate runtime, replace {self.__class__.__name__} "
+            f"with a runtime-specific operator class, possibly {self.__class__.__name__}Triton"
+        )
+
+    @classmethod
+    def from_model_registry(
```
karlhigley, line 458: I don't think I'm ready to say that model registry integration should be part of Merlin's core functionality. It sort of makes sense in Systems, but even there I wonder how directly it needs to be integrated into operator classes. If we're going to put an `export()` method in Core, then I do think it makes sense to have a corresponding method that loads it back from the same path (i.e. `from_path` or some variant of it), but this convenience method adds unnecessary coupling between operators and model registries.

merlin/dag/base_operator.py
```diff
@@ -409,3 +410,92 @@ def _get_columns(self, df, selector):
...
+            Other kwargs to pass to your Operator's constructor.
+
+        Returns
+        -------
+        InferenceOperator
+            New node for graph.
+        """
+        raise NotImplementedError(f"{cls.__name__} operators cannot be instantiated with a path.")
+
+    # TODO: This operator should be moved once all triton specific op migrations completed
```
karlhigley, line 498: Definitely drop this part, which I'm hoping we don't even need in Systems anymore

merlin/dag/utils.py
```diff
@@ -67,3 +71,61 @@ def group_values_offsets(flat_cols: dict) -> dict:
             grouped_cols[key] = value
 
     return grouped_cols
+
+
+class ModelRegistry(ABC):
```
karlhigley, line 76: I don't think this concept belongs in Core—it could maybe become the seed of the Merlin Ops/Orchestration library we've talked about for a while now?

karlhigley: @nv-alaiacano, do we have enough between the loose ends in Systems and the code/learning from AI Workflows to serve as a seed crystal for a new Merlin library?

merlin/dag/node.py
```diff
@@ -452,21 +452,22 @@ def export(
...
+            The base path to write this node's export directory.
         node_id : int, optional
             The id of this node in a larger graph (for disambiguation), by default None.
         version : int, optional
-            The Triton model version to use for this config export, by default 1.
+            The version of the node to use for this export, by default 1.
 
         Returns
         -------
         ModelConfig
```
karlhigley, line 469: I think `ModelConfig` is Triton-specific too

jperez999: So what should we put here? We dont know what it would be in any other case...

karlhigley: Or just remove the `Returns` section

jperez999: OK I removed the return all together as advised.

merlin/dag/base_operator.py
```diff
@@ -20,6 +20,7 @@
 from typing import Any, List, Optional, Union
 
 import merlin.dag
+import merlin.dag.utils
```
karlhigley, line 23: Unless I'm missing something, this import now seems unnecessary?

jperez999: good call

PR #358, opened 2023-07-06
move base runtime to core

jperez999:
This PR migrates the base runtime class to core. This is the first step in allowing the use of runtimes throughout merlin.

### DIFF
--- a/merlin/dag/base_runtime.py
+++ b/merlin/dag/base_runtime.py
@@ -1,0 +1,59 @@
+#
+# Copyright (c) 2022, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


merlin/dag/base_runtime.py
```diff
@@ -0,0 +1,59 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from merlin.core.protocols import Transformable
+from merlin.dag import Graph
+from merlin.dag.executors import LocalExecutor
+
+
+class Runtime:
+    """A Systems Graph Runtime.
```
karlhigley, line 22:
```suggestion
    """
    A Runtime provides operator implementations and an executor for running
    operator transforms in a specific environment (e.g. Dask, Triton.)
```

karlhigley (approved):
Approving since the code looks good, but the docstrings and copyright year could use an update before merging

PR #359, opened 2023-07-12
migrate base operator to operator and base_runtime to runtime

jperez999:
This PR migrates the Base_operator and base_runtime to operator and runtime for both the file names and the class names (where necessary). There are now two basic operator class types, operator and stat operator (which inherits from operator).


merlin/dag/__init__.py
```diff
@@ -14,9 +14,10 @@
 # limitations under the License.
 #
 
-# flake8: noqa
-from merlin.dag.base_operator import BaseOperator, DataFormats, Supports
 from merlin.dag.graph import Graph
 from merlin.dag.node import Node, iter_nodes, postorder_iter_nodes, preorder_iter_nodes
+
+# flake8: noqa
```
karlhigley, line 20: I think this should stay at the top

merlin/dag/operator.py
```diff
@@ -409,3 +411,43 @@ def _get_columns(self, df, selector):
...
+            The version of the operator, by default 1.
+
+        Returns
+        -------
+        model_config: dict
+            The config for the exported operator.
+        """
+
+
+BaseOperator = Operator
```
karlhigley, line 453: I'd probably put this in the init file, since the path to this file has changed and it will therefore break imports anyway

karlhigley (approved):

PR #364, opened 2024-01-12
use HAS_GPU to determine of cuda is available

jperez999:
This PR changes how we determine if cuda is available on the system. We move from numba to using HAS_GPU which uses nvml device count. If there are no devices, then cuda is not available. Otherwise cuda is available.

### DIFF
--- a/merlin/core/compat/__init__.py
+++ b/merlin/core/compat/__init__.py
@@ -101,7 +101,6 @@ def device_mem_size(kind="total", cpu=False):
     return pynvml_mem_size(kind=kind)
 
 
-
 try:
     import numpy
 except ImportError:
--- a/merlin/io/writer.py
+++ b/merlin/io/writer.py
@@ -196,7 +196,11 @@ class ThreadedWriter(Writer):
         if self.shuffle:
             df = shuffle_df(df)
         int_slice_size = df.shape[0] // self.num_out_files
-        slice_size = int_slice_size if int_slice_size > 0 and df.shape[0] % int_slice_size == 0 else int_slice_size + 1
+        slice_size = (
+            int_slice_size
+            if int_slice_size > 0 and df.shape[0] % int_slice_size == 0
+            else int_slice_size + 1
+        )
         for x in range(self.num_out_files):
             start = x * slice_size
             end = start + slice_size


merlin/core/compat/__init__.py
```diff
@@ -21,7 +21,7 @@
 
 from merlin.core.has_gpu import HAS_GPU  # noqa pylint: disable=unused-import
 
-if not cuda.is_available():
+if not HAS_GPU:
     cuda = None
```
rjzamora, line 25: Thanks @jperez999 !

I think we should also remove the use of `numba.cuda` below. In `device_mem_size`, we `try` to use `cuda.current_context().get_memory_info` to query the memory size. However, this creates a cuda context and returns exactly the same information as `pynvml_mem_size` anyway:


```python
    try:
        if kind == "free":
            return int(cuda.current_context().get_memory_info()[0])
        else:
            return int(cuda.current_context().get_memory_info()[1])
    except NotImplementedError:
        if kind == "free":
            # Not using NVML "free" memory, because it will not include RMM-managed memory
            warnings.warn("get_memory_info is not supported. Using total device memory from NVML.")
        size = pynvml_mem_size(kind="total", index=0)
        return size
```

Note that we were originally using `rmm.get_info` in the `try` block (I think), because that method used to tell us how much of the allocated memory pool was also free. However, rmm deprecated that method, because it became difficult to support. Using `pynvml` and `numba.cuda` will both the rmm pool allocation as "used" memory, and so there is no good reason to use `numba.cuda` over `pynvml` (in fact, it seems that it breaks `dask_cuda`!)

Disclaimer: I'm only 90% sure of what I'm saying. So, please feel free to push back if you think I'm wrong.

rjzamora: Ah, just realized @pentschev [already called out the same thing](https://github.com/NVIDIA-Merlin/core/pull/364#issuecomment-1888610944). Thanks Peter!

jperez999: @rjzamora  I have fixed this issue and removed cuda from use. We are still using numba in other parts of the merlin ecosystem so I wont completely remove it. But if its not here during initialization, we should be good to allow dask to create the context worry free.

rjzamora (approved):
Linting may be off but content looks good to me. Thanks @jperez999 !

jperez999:
This is not ready, the failures during writing have to do with when you are writing a file with a client available. Will continue investigating.

jperez999:
Investigated seems that the logic for int_slice_size was not full proof. Because of the floor divide you can find yourself in a scenario where you have less records in the df than the int_slice_size and that can result in a zero. Then when you go to mod on zero the thread raises an exception. I do wonder how we hit this now and not before.

rjzamora:
>I do wonder how we hit this now and not before.

I agree that this is strange - I wonder if I was wrong about `pynvml_mem_size` be "the same".

PR #374, opened 2024-07-19
Fix `test_hive_partitioned_data` for rapids 24.04

rjzamora:
Generalizes the type-casting operation in `test_hive_partitioned_data` to be version agnostic (needed for rapids 24.04).


tests/unit/io/test_io.py
```diff
@@ -190,7 +190,7 @@ def test_io_partitions_push(tmpdir):
     # Generate random csv files
     files = [os.path.join(tmpdir, f"csv/day_{i}") for i in range(23)]
     for file in files:
-        with open(file, "w") as f:
+        with open(file, "w", encoding="utf-8") as f:
```
rjzamora, line 193: This is just to make pylint happy.

jperez999 (approved):

